PR #4, opened 2017-08-13
GRU

Anonymission:
Adds working GRU layer and test, as well as a dot product operation


mygrad/nnet/layers/recurrent.py
```diff
@@ -158,3 +159,158 @@ def simple_RNN(X, U, W, s0=None, bp_lim=None):
...
+        s = self._hidden_seq.data[:-1]
+        z = self._z.data
+        r = self._r.data
+        h = self._h.data
+
+        dLds = grad[1:]
+
+        for i in range(min(s.shape[0] - 1, self.bp_lim)):
+            #  dL(t) / ds(t) + dL(t+1) / ds(t)
+            dLds[len(dLds) - (i + 2)] += _gru_dLds(s[len(dLds) - (i + 1)],
```
LLrsokl: I don't think that this does truncated backprop through time correctly.  Note that the simple-rnn updates a **slice** of the dLds array:

```python
dLt_dst[1:len(grad) - (i + 1)] += ...
```

The whole reason I had to do the numpy-copy and array-diff step:

```python
index = slice(2, len(grad) - i)
dLn_ft1 = dst_dft[index] * (dLt_dst[index] - old_dst[index])
old_dst = np.copy(dLt_dst)
```
was to perform truncated back-prop properly. Each hidden descriptor needs to pass its current gradient to the left.

Currently, your code only works if you backprop through the whole sequence. If `bp_lim=2` then only the *final* hidden-descriptor would backprop its gradient back by two steps. This isn't caught by the unit-test because you can't enforce the truncated backprop in the recursive form of the GRU. 

I think this should be a pretty simple fix, using simple-rnn as the template. What do you think?

mygrad/math.py
```diff
@@ -115,3 +115,25 @@ def subtract(a, b):
         -------
         mygrad.Tensor"""
     return Tensor._op(Subtract, a, b)
+
```
LLrsokl: Can you leave the `dot` and `Dot` (operation) changes out of this PR? Let's have it just be of the GRU, and discuss the `dot` operation separately.

My only hang up with it is that people expect `dot` to behave like numpy-dot which covers both vector dot products and matrix multiplication - the latter is covered by the `dense` layer. So I just want to think about how to deal with this design issue before committing.

mygrad/nnet/layers/recurrent.py
Anonymission: I went through the math again and agree. I was having problems with the simple-rnn slice approach earlier, but just implemented a fix and it is now working properly (slicing was an index off for the s, z, r, and h parameters to _gru_dLds)

mygrad/nnet/layers/recurrent.py
```diff
@@ -158,3 +159,170 @@ def simple_RNN(X, U, W, s0=None, bp_lim=None):
...
+        self.Wz.backward(np.einsum("ijk, ijl -> kl", s, dz))
+        self.bz.backward(np.sum(np.sum(dz, axis=0), axis=0))
+
+        self.Ur.backward(np.einsum("ijk, ijl -> kl", self._input_seq, dr))
+        self.Wr.backward(np.einsum("ijk, ijl -> kl", s, dr))
+        self.br.backward(np.sum(np.sum(dr, axis=0), axis=0))
+
+        self.Uh.backward(np.einsum("ijk, ijl -> kl", self._input_seq, dh))
+        self.Wh.backward(np.einsum("ijk, ijl -> kl", (s * r), dh))
+        self.bh.backward(np.sum(np.sum(dh, axis=0), axis=0))
```
LLrsokl: I just realized that we need to backprop through `X` too! In case you chain GRUs together (i.e. the hidden descriptors of the first GRU is `X` for the subsequent)

Anonymission: Should now backdrop through x correctly, assuming that partial ds_t-1/dx does not play a role in the differentiation of ds_t/d_x

LLrsokl:
@Anonymission I pulled this into master on my fork of MyGrad. There were a couple of issues with null_gradients that I fixed and added a couple of checks to the unit test. This is so awesome! Excellent work. I'm really interested to see this be trained.

Anonymission:
Great! I need to set up a spare computer to let this train for a while, but once it's done I'll let you know the results.

On a side note, the _gru_dLdx function is not required for the code to work; it was part of my experimenting and I forgot to remove it.

PR #5, opened 2017-08-16
Optim

Anonymission:
Numba optimizations for simple_rnn and gru __call__ and backward functions.

When measured with line_profiler, backpropping the gru takes about 1/2 the time as the current non-numba version on a small sequence of data (after numba has compiled).


LLrsokl:
@Anonymission I just saw this! Not sure how I overlooked it. These look like some great optimizations.

I'll have to resolve conflicts with GRU update I just made with dropout. 

Looks like you made the deep dive into numba ;)

PR #60, opened 2018-06-17
Network Graphs

Anonymission:
Adds network graphs using ~~pydot and~~ graphviz

### DIFF
--- a/mygrad/computational_graph.py
+++ b/mygrad/computational_graph.py
@@ -1,14 +1,29 @@
 import pydot
+from mygrad.tensor_base import Tensor
 
-
-def build_graph(fin):
-    """ Builds and saves a network graph as a PNG file.
+def build_graph(fin, names=None):
+    """ Builds and saves a computational graph as a PNG file.
 
         Parameters
         ----------
         fin : Tensor
-            The tensor object that should be the final node in the
+            The tensor object that will be the final node in the
             computational graph.
+
+        names : Optional[Dict[str, Tensor]]
+            A dictionary that maps names of Tensors to Tensor objects. If
+            an argument is passed to names, the key name that maps to a Tensor
+            included in the computational graph will be used as a label for the
+            Tensor's node.
+
+            To use the names assigned in the local environment,
+            pass `names=locals()` to the build_graph function.
+
+            If different names are used from the local environment,
+            the key must map to the exact Tensor object. A new Tensor or copy
+            of the original Tensor should not be created as the value in the
+            dictionary.
+
 
         Notes
         -----
@@ -16,6 +31,9 @@ def build_graph(fin):
 
         The PNG file will be written to the current working directory.
     """
+    assert isinstance(fin, Tensor), "fin must be a Tensor"
+    assert isinstance(names, dict), "names must be a dictionary"
+
     graph = pydot.Dot(graph_type='digraph')
 
     for out, op in fin._graph_dict.items():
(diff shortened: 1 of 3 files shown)


mygrad/computational_graph.py
```diff
@@ -0,0 +1,107 @@
...
+    node_id = str(id(node))
+
+    graph.node(name=node_id, label=node_lab)
+
+    if node._creator is None:
+        if op_id is not None:
+            graph.edge(node_id, op_id)
+        return
+    else:
+        op_lab = node._creator.__repr__().rpartition(".")[-1].split(" ")[0]
```
rsokl: You should call `repr` rather than invoking the dunder method `.__repr__()`. These accomplish the same thing, but the former is more readable.

```python
op_lab = repr(node._creator).rpartition(".")[-1].split(" ")[0]
```

mygrad/tensor_base.py
```diff
@@ -277,7 +277,7 @@ def creator(self):
 
     def __len__(self):
         return len(self.data)
-
```
rsokl: In the future, if you accidentally "touch" files like this, just do `git checkout -- <file>` to undo the modifications. This will ensure that only the relevant files are affected by your PR.

mygrad/computational_graph.py
```diff
@@ -0,0 +1,107 @@
+from graphviz import Digraph
+from mygrad.tensor_base import Tensor
+import numpy as np
+
+def build_graph(fin, names=None, render=True, save=False, dims=False, dtypes=False, sum_stats=False):
+    """ Builds and renders a computational graph.
+
+        Parameters
+        ----------
+        fin : Tensor
```
rsokl: type-hint this as `mygrad.Tensor`

mygrad/computational_graph.py
```diff
@@ -0,0 +1,107 @@
...
+    if save:
+        graph.render(filename="computational_graph", cleanup=True)
+
+    if render:
+        return graph
+
+
+def _add_node(node, graph, op_id=None, **kwargs):
+    """ Recursively traces computational graph and adds nodes to Digraph. """
+    node_lab = node.__repr__()
```
rsokl: --> `repr(node)`

mygrad/computational_graph.py
```diff
@@ -0,0 +1,107 @@
...
+    if kwargs['names'] is not None:
+        for key in kwargs['names']:
+            if id(kwargs['names'][key]) == id(node):
+                node_lab = key
+    if kwargs['dims']:
+        node_lab = node_lab + "\nDims: " + str(node.shape)
```
rsokl:
```python
"\nDims: " + str(node.shape)
```
is better written as:
```python
"\nDims: {}".format(node.shape)
```
This is especially the case for longer strings. You can do named formatted fields:

```python
"{apple}, {moo}, {cat}".format(apple=2, moo="hi", cat=True)
```

Anonymission:
I uninstalled graphviz, and importing mygrad normally still works. An error only is thrown when I try to import `build_graph`. 

Here are the few scenarios related to rendering a graph in-notebook:

```python
from mygrad.computational_graph import build_graph

a = Tensor(np.array([[1, 2], [3, 4]]))
b = Tensor(np.array([[2, 3], [9, 7]]))
c = a + b

# calling build_graph without assigning the return value to a variable 
# will render the graph automatically
build_graph(c, names=locals())
# RENDERS GRAPH

# assigning the output to a variable will store the graphviz.Digraph object, 
# which will render if called
graph = build_graph(c, names=locals())
graph
# RENDERS GRAPH

# passing render=False to build_graph will not return the Digraph object, 
# thus not rendering the graph in-notebook
build_graph(c, names=locals(), render=False)
# DOES NOT RENDER GRAPH
```

mygrad/tensor_base.py
Anonymission: I can still revert these changes since they're isolated to `tensor_base.py` and I removed all the other changes I previously had in that file. Want me to undo these?

Anonymission:
I'll need to update this so that if `names=locals()` is passed, it also is able to find the names of any non-Tensor object that was later converted to a Tensor (such as if a ndarray and Tensor are added together).

PR #61, opened 2018-06-17
Tensor joining

Anonymission:
Adds ops, functions, and tests for tensor joining.

Includes:
- concatenate
- stack
- hstack
- vstack
- dstack


Anonymission:
Commit 24095ec accidentally changed every file

PR #75, opened 2018-07-21
add 1d, 2d, 3d conv unit tests

rsokl:


rsokl:
@Anonymission we should be in good shape now :)

We'll need the kids to pull mygrad for the fix

Issue #58, opened 2018-06-16
argmax/argmin


davidmascharka:
They should exist as non-differentiable functions in Tensor. These should merely be wrappers on the numpy functions so that they can also accept tensors. These functions should return numpy arrays.

alexding123:
Would like to handle this. Not a contributor so can't self-assign. Will pull-request when done.

davidmascharka:
Awesome! I'm not sure how to best handle assignment since this isn't an org so for now I'll make a new "in-progress" tag.

@rsokl can chime in with how he wants the operations separated out (i.e. in a `nondifferentiable` package or subsumed under the current `math` package).

rsokl:
@alexding123 I think that these should go under `mygrad.math.nondifferentiable` (of course they'll be available at top-level, so people wont have to know how to spell that). These will exist as functions, which we can then simply set as methods

We'll want to forego calling `_op` as the output of argmax/argmin will be numpy-arrays and not tensors, and there is no need to hook into any computational graph at all. Thus you will simply want to implement these as functions that normalize the input to numpy's argmax/argmin functions so that mygrad-tensors can be accepted in addition to other array_like objects.

Once you write those function, we can still conform to the pattern of making these functions as methods of `Tensor` in addition to being standalone functions, as seen [here](https://github.com/rsokl/MyGrad/blob/d6dff60f598a3c28467832e60043187af4c7db3c/mygrad/__init__.py#L19). This will make maintainability and testability simple, as you were hoping for.

Let me know if this makes sense.

PR #86, opened 2018-09-04
Argmin/max and test files

alexding123:
See the changes made as outlined [here](https://github.com/rsokl/MyGrad/issues/58#issue-333017179). I did not support numpy's `array_like` in my implementation since `mygrad.argmin` and `mygrad.argmax` are mainly used as methods, and users should simply opt for numpy's methods to do the same operations to non-Tensors. I ran my own test codes and did some simple manual testing. Everything seems to be in working order. 

Also, added mac compatibility to the gitignore file.

### DIFF
--- a/tests/math/unary/test_nondifferentiable.py
+++ b/tests/math/unary/test_nondifferentiable.py
@@ -1,4 +1,4 @@
-from mygrad import argmin, argmax
+from mygrad as mg
 from mygrad.tensor_base import Tensor
 
 import hypothesis.strategies as st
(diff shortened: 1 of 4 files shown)


rsokl:
MyGrad is designed such that array-like objects can be used interchangeably with constant-tensors. The `argmax`/`argmin` functions shouldn't behave any differently. You can simply use:

```python
a = a.data if isinstance(a, Tensor) else np.asarray(a)
```
to normalize the input. 

As it is, your functions inadvertently support numpy-arrays as-is (try it out  😛), but fails with a cryptic error for other array_like objects:

```python
>>> import mygrad as mg
>>> mg.argmax([1, 2, 3])
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-16-83de536c8caa> in <module>()
----> 1 mg.argmax([1, 2, 3].data)

AttributeError: 'list' object has no attribute 'data'
```

alexding123:
Just updated the pull request!

tests/math/unary/test_nondifferentiable.py
```diff
@@ -0,0 +1,25 @@
...
+import numpy as np
+from hypothesis import given
+
+import numpy as np
+
+@given(d=st.integers(min_value=3, max_value=10))
+def test_argmin(d):
+    dimensions = tuple(np.random.randint(1,6) for i in range(d)) 
+
+    a = np.random.randint(-100,100, size=dimensions)
```
davidmascharka: This would be good to do over other data types as well, but Ryan can chime in with testing

rsokl: Yes, you can use the following hypothesis strategy to test various dtypes:

```python
dtype_strat_numpy = st.sampled_from((np.int8, np.int16, np.int32, np.int64,
                                     np.float16, np.float32, np.float64))
```

Additionally, you should use `hypothesis` to draw the arrays, rather than use numpy to generate them randomly. You can also use mygrad's custom search strategy `valid_axes` - this will generate the appropriate positive and negative axis values, in addition to `None`

```python
from ...custom_strategies import valid_axes

@given(a=hnp.arrays(shape=hnp.array_shapes(max_side=4, max_dims=5),
                    dtype=dtype_strat_numpy ,
                    elements=st.floats(-1e-6., 1.)),
       data=st.data())
def test_transpose(a, data):
    axis = data.draw(valid_axes(ndim=a.ndim, single_axis_only=True), label="axis")
```

Using the `st.data()` search strategy allows you  draw axes based on the array `a` that was generated via the `@given` decorator.

This is far superior to randomly drawing an array as:
- hypothesis does not merely generate random numbers, it specifically generates elements to find edge cases that break code. It also simplifies the fail case whenever possible before reporting it.
- you cannot recreate or even see the random array that happened to break your unit test. Hypothesis will nicely report all of the objects that it generated if the test fails

rsokl: Whoops! Good catch. Thanks.
(edited in the earlier post to fix this)

alexding123: I updated the code. One problem with the code you wrote though: it generates Float regardless of what the sampled dtype is, which results in Hypothesis complaining about downsampling when it tries to convert the Float into, say, Int, when Int is the sampled dtype. I just left the `element` parameter as `None`.

alexding123:
Made the changes as requested.

Issue #87, opened 2018-10-21
Missing math functions


rsokl:
The following math functions need implementations in `mygrad.math`. Please refer to the contents of that module for implementations of functions like `mygrad.exp` , `mygrad.sqrt`, `mygrad.arctan` for some guidance. 

- [exp2](https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.exp2.html)
- [square](https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.square.html)
- [arctan2](https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.arctan2.html)

IanCoolidge0:
I'll pull request these three tonight. Is there any special optimization I should do for the square function or is it just there to mimic numpy?

davidmascharka:
Addressed in #88

PR #88, opened 2018-10-21
add square, exp2, and arctan2

IanCoolidge0:
Here are the three functions from #87.

I'm not entirely sure what I'm doing with the unit tests so they are mostly copied and pasted - nothing happened when I ran them so I assume that's a good thing.


rsokl:
This looks great! I'll get to reviewing this carefully ASAP

IanCoolidge0:
Turns out I forgot to write the arctan2 unit test, it should work now. Is there a reason that the existing trig functions don't have a constant parameter? I needed to add a constant parameter to `arctan2` to make the unit tests work

PR #95, opened 2018-11-12
matmul

AFederici:
Made a small test for it and it seems to work.  Its pretty derivative of what NumPy had.  I changed a lot of the wording to fit matmul instead of dot, got rid of certain NumPy functions and implemented them directly, and changed around the indexing in order for this to work for N-D instead of 2-D.

One thing I was unsure of was lines 87-93.  In numpy's implementation it made sense dimensions could've been explicitly added, but is this something we need to worry about for matmul?  I didn't put much thought into it as I assumed the constant keyword would deal with the first case and the second case involving calling ravel() was throwing an error so I just commented out the whole section


rsokl:
@AFederici this is great! Thanks so much for taking the time to put this together. I made some minor adjustments to take care of the edge cases where the end-tensors are 1D and wrote a unit test.

PR #128, opened 2019-06-17
Add `tox -e format` and apply all the formatting tools

Zac-HD:
You... probably want to review this one on a per-commit basis, and might just take my word for the automated ones ("Apply $tool").  In short:

- Configured a `format` environment for `tox`, which will apply `autoflake`, `isort`, and `black`.
- Applied each tool with a separate commit, for easier review.
- Also applied `pyformat --py3only $(find . -name '*.py')`; that's not automated though as the author doesn't want to support directories and the find-invocation is a little fragile (on e.g. Windows)
- Did some manual cleanup after applying Black and reviewed all diffs.

I didn't make this run in CI, but that's easy enough to add later :smile:


rsokl:
Thanks @Zac-HD ! This was a much needed makeover for the codebase! This will be quite the peak in our repo-insights plots 😆

Issue #133, opened 2019-07-17
Add mygrad.pi and other convenient constants


rsokl:

rsokl:
Closed by #140

PR #140, opened 2019-08-08
Constants

mkhan45:
I kind of arbitrarily chose some constants after reading #133 and put them under `mygrad.math.consts`.

I added documentation using `automodule:: mygrad.math.consts` but it looks out of place.

documentation

Ironically I dislike the documentation for sphinx

### DIFF
--- a/mygrad/math/consts.py
+++ b/mygrad/math/consts.py
@@ -1,17 +1,35 @@
 import numpy as np
 
-Inf = np.array(np.Inf)
-Infinity = np.array(np.Infinity)
-NAN = np.array(np.NAN)
-NINF = np.array(np.NINF)
-NZERO = np.array(np.NZERO)
-NaN = np.array(np.NaN)
-PINF = np.array(np.PINF)
-PZERO = np.array(np.PZERO)
-e = np.array(np.e)
-euler_gamma = np.array(np.euler_gamma)
-inf = np.array(np.inf)
-infty = np.array(np.infty)
-nan = np.array(np.nan)
-newaxis = np.array(np.newaxis)
-pi = np.array(np.pi)
+__all__ = [
+    'Inf',
+    'Infinity',
+    'NAN',
+    'NINF',
+    'NZERO',
+    'NaN',
+    'PINF',
+    'PZERO',
+    'e',
+    'euler_gamma',
+    'inf',
+    'infty',
+    'nan',
+    'newaxis',
+    'pi',
+]
+
+Inf = np.Inf
+Infinity = np.Infinity
+NAN = np.NAN
+NINF = np.NINF
+NZERO = np.NZERO
+NaN = np.NaN
+PINF = np.PINF
+PZERO = np.PZERO
+e = np.e
+euler_gamma = np.euler_gamma
+inf = np.inf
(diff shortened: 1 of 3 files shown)


rsokl:
Thanks for the submission @mkhan45 ! 

A couple of notes, first of all the failing unit test is not due to your submission - we have a "flaky" test that needs fixing.

Next, mygrad generally strives to mimic numpy very closely in its content and structure (we basically want to be able to replace `import numpy` with `import mygrad` and have things "just work"). Thus, we should look to mirror [the constants that numpy implements](https://docs.scipy.org/doc/numpy/reference/constants.html). I am a big fan of the physical constants that you introduced, but let's just stick to those that numpy provides 😄 

Don't worry about building the docs for this, I will handle the sphinx stuff.

It may be best for you to make a pass at this in a new branch, in light of the feedback above. The next step will be to add some unit tests!

mkhan45:
@rsokl I was wondering why it passed for python 3.5 and 3.7 but not 3.6...

I changed everything to be the same as numpy and return as a numpy array but I did it in the same branch because I didn't read all the way through oops. I reset all the sphinx stuff though.

What kind of unit tests would there be for this?

mygrad/math/consts.py
```diff
@@ -0,0 +1,17 @@
+import numpy as np
+
+Inf = np.array(np.Inf)
```
rsokl: No need to cast these any of these as arrays, you can just make the alias:

```suggestion
Inf = np.Inf
```

MyGrad can work with float's just fine :)

mygrad/math/consts.py
```diff
@@ -0,0 +1,17 @@
+import numpy as np
+
```
rsokl: Include an `__all__`, which is a list of all of the names of the constants:

```python
__all__ = ["Infinity", "NAN", "NINF", ...
```

mygrad/math/consts.py
```diff
@@ -0,0 +1,17 @@
...
+NZERO = np.array(np.NZERO)
+NaN = np.array(np.NaN)
+PINF = np.array(np.PINF)
+PZERO = np.array(np.PZERO)
+e = np.array(np.e)
+euler_gamma = np.array(np.euler_gamma)
+inf = np.array(np.inf)
+infty = np.array(np.infty)
+nan = np.array(np.nan)
+newaxis = np.array(np.newaxis)
```
rsokl: FYI: `np.newaxis` is actually just an alias for `None` (I know, weird, right?). So we most definitely do not want to make it an array!

rsokl:
🎆 🎉 Thank @mkhan45 !!

mkhan45:
Yay my first contribution that's more than 2 characters. One question though:

Why are the tests necessary? Isn't it like declaring `x = 1` and then doing `assert x is 1`? The only way I could see it fail is if there was a typo but in that case there might be a typo in the tests too.

rsokl:
It serves as a nice regression test - who is to say that I, or another contributor, won't accidentally delete/change a line in that code while making edits to MyGrad? It is much better to have an automated test catch that than an unfortunate user who hits a bug!

PR #172, opened 2019-12-15
fix unused argument in backprop_test_factory; fix flaky test coverage

rsokl:


mygrad/math/sequential/ops.py
```diff
@@ -212,6 +212,7 @@ def backward_var(self, grad, index, **kwargs):
         # the nans in the sequences can simply be set to 0
         if np.any(np.isnan(dldx)):
             x = x.copy()
+            print("HERE")
```
aslvrstn: I don't think you meant to merge these.

rsokl: Oh crap...

PR #175, opened 2019-12-19
Reorganize into src layout

### LINKED ISSUE #168: Separate source code from setup.py
Move mygrad/ -> src/mygrad

davidmascharka:
Not sure if I need to update anything on the docs side

Closes #168


rsokl:
@davidmascharka sorry for the flaky coverage, can you update [this line](https://github.com/rsokl/MyGrad/blob/c9f06f74468f8f34668e8a70879213bbda876e28/tests/math/unary/test_scipy_mirror.py#L12)

```python
@settings(deadline=None)
```

to


```python
@settings(deadline=None, max_examples=500)
```

That should take care of the slippage.

rsokl:
@Zac-HD for reference, the following test:

https://github.com/rsokl/MyGrad/blob/c9f06f74468f8f34668e8a70879213bbda876e28/tests/math/unary/test_scipy_mirror.py#L12-L31

stopped covering [the last line](https://github.com/rsokl/MyGrad/blob/5ea7b929108c90a83c37c3d619cef0a6f6596e28/mygrad/math/_special.py#L55) in:

```python
    a = np.asarray(a)
    a_max = np.amax(a, axis=axis, keepdims=True)


    if a_max.ndim > 0:
        a_max[~np.isfinite(a_max)] = 0
    elif not np.isfinite(a_max):
        a_max = 0
```

Thus it seems like hypothesis never generated inf among any of the 100 arrays.

rsokl:
Ha. That time the job simply died... re-running.

rsokl:
@davidmascharka I don't think anything needs to be changed on the docs end - as long as `mygrad` can be installed / imported in the same way, RTD + sphinx should be happy :)

PR #181, opened 2020-01-09
Rename logsoftmax -> log_softmax and ReLu -> ReLU

davidmascharka:
I know ReLU doesn't change anything but.... Rectified Linear Unit

`logsoftmax` -> `log_softmax` is self-explanatory :yum: also will make it consistent with MyNN so this is the first step of the merge.


davidmascharka:
Closing this and opening an issue to do it since this would require a major version bump

PR #183, opened 2020-01-09
Reorganize nnet

davidmascharka:
To continue some setup for our merging, this does a little cleanup to avoid massive files down the road. Splits off `nnet.activations` and `nnet.losses` from monolithic files into modules (functionality and access are unchanged). Also performs this reorganization for `tests/nnet`. I think this will be a lot cleaner when we start adding all the losses and activations from MyNN.

Related: I'm not sure how you want the MyNN merge to go. The two ways I'm imagining are:

1. A component from MyNN gets moved into a branch and tests are added to maintain 100%, then that is merged into MyGrad. For example, I add tests for ELU and then it gets merged straight in.

2. Everything from MyNN gets merged into MyGrad as a single commit. Everything is moved over into a working branch and tests are added, once every piece of functionality is covered and moved over, there's a single giant merge. This keeps the git log super clean on what exactly happened but takes a lot longer.

If you prefer 2, let's just move the base of this from master to mynn-merging and we can go with that. Otherwise I'll kill that branch and start on strategy 1.


rsokl:
@davidmascharka excited to get this moving! I am happy with option 1.

One thing that we need to take care about is keeping the sphinx docs up to date, whether it be change where sphinx is importing from or adding materials for new functions like ELU

rsokl:
@davidmascharka in general, I will leave it to you to merge once I've left approval - just to make sure everything is final on your end.

davidmascharka:
:+1: I'll definitely mess around with sphinx on my end and see what works well. Not sure how you want to handle releases of MyGrad. A new release and minor version bump every time there's a new function merged over would probably get pretty annoying and we'd be onto MyGrad v1.93 pretty quickly :thinking:

rsokl:
> A new release and minor version bump every time there's a new function merged over would probably get pretty annoying.

I actually don't really mind that. Hypothesis has totally desensitized large minor version numbers. That being said, we need to be careful to make sure that any API-changing changes gets an appropriate major-bump. There is a good amount of tutorial/course material that will be broken by it.

We must take care to maintain a detailed [changelog](https://github.com/rsokl/MyGrad/blob/master/docs/source/changes.rst). Perhaps you can add changes to this file for this PR 

I think it would make a lot of sense for me to make you and admin on this repo and on the PYPI page. Are you okay with this?

davidmascharka:
That sounds totally reasonable. Updated the changelog for this.

PR #187, opened 2020-01-11
ensure that (non-array) floats are tested as input during forward pass

rsokl:


tests/wrappers/uber.py
```diff
@@ -119,6 +120,9 @@ def __init__(
             A callable that is fed the generated arrays and keyword arguments that will
             be fed to ``mygrad_func``. If ``assumptions`` returns ``False``, that test
             case will be marked as skipped by hypothesis.
+
+        permit_0d_array_as_float : bool, optional (default=True)
+            If True, drawn 0D arrays will potentially be case to numpy-floats.
```
davidmascharka:
```suggestion
            If True, drawn 0D arrays will potentially be cast to numpy-floats.
```

tests/wrappers/uber.py
```diff
@@ -243,6 +248,16 @@ def wrapper(shapes: hnp.BroadcastableShapes, constant, data: st.DataObject):
                 for value in self.index_to_no_go.get(i, ()):
                     assume(np.all(arr != value))
 
+            if self.permit_0d_array_as_float:
+                # potentially cast a 0D array as a float
+                arrs = tuple(
+                    arr.item()
+                    if arr.ndim == 0
+                    and data.draw(st.booleans(), label="arr-{} to float".format(n))
```
davidmascharka: is there a plan to drop 3.5?

rsokl: Not until it officially hits EOL :/

rsokl:
@davidmascharka should be all set now

PR #192, opened 2020-01-20
Allow y_true to be a Tensor in softmax-crossentropy

rsokl:


src/mygrad/nnet/losses/softmax_crossentropy.py
```diff
@@ -26,6 +26,8 @@ def __call__(self, x, y_true):
             Returns
             -------
             The average softmax loss"""
+        if isinstance(y_true, Tensor):
+            y_true = y_true.data
```
davidmascharka: should we have a function like `np.asarray` that just does this type of check? e.g.:

``` python
def asarray(x):
    if isinstance(x, Tensor):
        x = x.data
    return np.asarray(x)
```

to handle this? We're bound to be doing this check a lot

tests/nnet/losses/test_softmaxcrossentropy.py
```diff
@@ -27,31 +27,30 @@ def test_input_validation(data, labels):
...
     scores = Tensor(s)
-    softmax_cross = softmax_crossentropy(scores, loss, constant=False)
+    softmax_cross = softmax_crossentropy(scores, y_true, constant=False)
     softmax_cross.backward()
 
     mygrad_scores = Tensor(s)
     probs = softmax(mygrad_scores)
 
-    correct_labels = (range(len(loss)), loss)
+    correct_labels = (range(len(y_true)), y_true.data if labels_as_tensor else y_true)
```
davidmascharka:
```suggestion
    correct_labels = (range(len(y_true)), y_true.data if not labels_as_tensor else y_true)
```

?

rsokl: If the labels are tensors then we need to access their underlying numpy array

PR #196, opened 2020-01-24
Fix softmax

### LINKED ISSUE #193: softmax does not work on integer-type arrays
image

davidmascharka:
Fixes #193 

will add a test for this. The softmax tests are 🤨

### DIFF
--- a/src/mygrad/nnet/activations/softmax.py
+++ b/src/mygrad/nnet/activations/softmax.py
@@ -7,7 +7,8 @@ from mygrad.tensor_base import Tensor
 
 def _softmax(x, kwargs):
     x = x - x.max(**kwargs)
-    x = x.astype(np.float)
+    if np.issubdtype(x.dtype, np.integer):
+        x = x.astype(np.float)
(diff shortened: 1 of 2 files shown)


src/mygrad/nnet/activations/softmax.py
```diff
@@ -7,6 +7,7 @@
 
 def _softmax(x, kwargs):
     x = x - x.max(**kwargs)
+    x = x.astype(np.float)
```
rsokl: We should cast to float only if x is an int-array. Otherwise if the user inputs `float32`, this will cast to `float64`

rsokl:
Yeah, it's ironic because I just wrote some nice property-based tests for softmax in my testing tutorial... The whole nnet section needs a testing overhaul (except conv and gru - I think those are solid).

Also I need to update my hypothesis tests to include int values to catch stuff like this :/

rsokl:
*sigh* currently the test factories only pass floats through the functions - so you'll need to write a simple test that verifies that floats work ok (just check forward pass)

PR #205, opened 2020-02-15
adds Tensor.astype

### LINKED ISSUE #115: Add astype method to Tensor
How don't we have this?? [Numpy ref](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.astype.html)

rsokl:
Closes #115

### DIFF
--- a/src/mygrad/tensor_base.py
+++ b/src/mygrad/tensor_base.py
@@ -5,7 +5,7 @@ etc., are bound to the Tensor class in ``mygrad.__init__.py``.
 """
 
 from functools import wraps
-from typing import Set, Type, Union
+from typing import Optional, Set, Type, Union
 
 import numpy as np
 
@@ -157,6 +157,31 @@ class Tensor:
 
         # track the operations that have contributed to this tensor's gradient during a back-prop
         self._accum_ops = set()  # type: Set[Operation]
+
+    def astype(self, dtype: type, *, constant: Optional[bool] = None) -> "Tensor":
+        """Returns a distinct tensor with its data modified to have the speified
+        data type.
+
+        The resulting tensor does belong to any pre-existing computation graph; i.e.
+        it is as if this tensor was created 'by scratch'.
+
+        Parameters
+        ----------
+        dtype : type
+            The real-values data type
+
+        constant : Optional[bool]
+            If specified, determines in the returned tensor is a constant.
+            Otherwise this argument is inferred from the original tensor.
+
+        Returns
+        -------
+        Tensor
+            The resulting tensor with the specified data type.
+        """
+        constant = constant if constant is not None else self.constant
+        self._check_valid_dtype(dtype)
+        return type(self)(self.data.astype(dtype), constant=constant)
 
     @staticmethod
     def _check_valid_dtype(dtype):
--- a/src/mygrad/tensor_base.py
+++ b/src/mygrad/tensor_base.py
@@ -159,7 +159,7 @@ class Tensor:
         self._accum_ops = set()  # type: Set[Operation]
 
     def astype(self, dtype: type, *, constant: Optional[bool] = None) -> "Tensor":
-        """Returns a distinct tensor with its data modified to have the speified
+        """Returns a distinct tensor with its data modified to have the specified
         data type.
 
         The resulting tensor does belong to any pre-existing computation graph; i.e.
@@ -171,7 +171,7 @@ class Tensor:
             The real-values data type
 
         constant : Optional[bool]
-            If specified, determines in the returned tensor is a constant.
+            If specified, determines if the returned tensor is a constant.
             Otherwise this argument is inferred from the original tensor.
 
         Returns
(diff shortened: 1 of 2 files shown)


src/mygrad/tensor_base.py
```diff
@@ -158,6 +158,30 @@ def __init__(
         # track the operations that have contributed to this tensor's gradient during a back-prop
         self._accum_ops = set()  # type: Set[Operation]
 
+    def astype(self, dtype: type, *, constant: Optional[bool] = None) -> "Tensor":
+        """Returns a distinct tensor with its data modified to have the speified
```
aslvrstn: s/speified/specified/

src/mygrad/tensor_base.py
```diff
@@ -158,6 +158,30 @@ def __init__(
...
+        The resulting tensor does belong to any pre-existing computation graph; i.e.
+        it is as if this tensor was created 'from scratch'.
+
+        Parameters
+        ----------
+        dtype : type
+            The real-values data type
+
+        constant : Optional[bool]
+            If specified, determines in the returned tensor is a constant.
```
aslvrstn: s/in/if/

rsokl:
@davidmascharka mind giving this a look?

src/mygrad/tensor_base.py
```diff
@@ -158,6 +158,30 @@ def __init__(
         # track the operations that have contributed to this tensor's gradient during a back-prop
         self._accum_ops = set()  # type: Set[Operation]
 
+    def astype(self, dtype: type, *, constant: Optional[bool] = None) -> "Tensor":
+        """Returns a distinct tensor with its data modified to have the specified
+        data type.
+
+        The resulting tensor does belong to any pre-existing computation graph; i.e.
```
davidmascharka:
```suggestion
        The resulting tensor does not belong to any pre-existing computation graph; i.e.
```

tests/tensor_base/test_astype.py
```diff
@@ -0,0 +1,33 @@
...
+import hypothesis.strategies as st
+
+from mygrad import Tensor
+from numpy.testing import assert_array_equal
+
+
+@given(
+    tensor=st.tuples(
+        hnp.arrays(
+            shape=hnp.array_shapes(), dtype=hnp.integer_dtypes() | hnp.floating_dtypes()
```
davidmascharka: does this cover uints or just signed ints?

tests/tensor_base/test_astype.py
```diff
@@ -0,0 +1,33 @@
...
+        st.booleans(),
+    ).map(lambda x: Tensor(x[0], constant=x[1])),
+    dest_type=hnp.integer_dtypes() | hnp.floating_dtypes(),
+    constant=st.booleans() | st.none(),
+)
+def test_astype(tensor: Tensor, dest_type: type, constant: Optional[bool]):
+    tensor = tensor * 1  # give tensor a creator
+    new_tensor = tensor.astype(dest_type, constant=constant)
+
+    assert new_tensor.constant is (tensor.constant if constant is None else constant)
```
davidmascharka: not sure which of these I like better :thinking: 

```suggestion
    assert new_tensor.constant is (constant or tensor.constant)
```

tests/tensor_base/test_astype.py
```diff
@@ -0,0 +1,33 @@
...
+    assert tensor.creator is not None
+    assert new_tensor.creator is None
+    assert new_tensor.dtype is dest_type
+    assert new_tensor.shape == tensor.shape
+
+    if new_tensor.dtype is tensor.dtype:
+        assert_array_equal(new_tensor.data, tensor.data)
```
davidmascharka: do you think it's worthwhile to add a test that round-trips from a lower-precision dtype to a higher-precision dtype and back and verifies the data is unchanged? e.g. float16 -> float32 -> float16 and assert_array_equal(new_tensor.data, tensor.data)` or int16/32 -> int16 -> int16/32

tests/tensor_base/test_astype.py
rsokl: The former is necessary because I only want to defer to `tensor.constant` if `constant` was `None`. If `constant` is explicitly set as `False` then that should be used.

tests/tensor_base/test_astype.py
rsokl: *sigh*... yes that is worthwhile 😢 

Like.. we should definitely make sure that we don't change people's data on them.... I'll add this.

tests/tensor_base/test_astype.py
rsokl: Yeah it is just signed ints, I can add unsigned

tests/tensor_base/test_astype.py
```diff
@@ -0,0 +1,59 @@
...
+@pytest.mark.parametrize(
+    "type_strategy",
+    [hnp.integer_dtypes(), hnp.unsigned_integer_dtypes(), hnp.floating_dtypes()],
+)
+@given(data=st.data())
+def test_upcast_roundtrip(type_strategy, data: st.DataObject):
+    thin, wide = data.draw(
+        st.tuples(type_strategy, type_strategy).map(
+            lambda x: sorted(x, key=lambda y: np.dtype(y).itemsize)
+        )
```
davidmascharka: ooooo nice

rsokl: 😎

PR #214, opened 2020-03-01
Add type hints

rsokl:
itshappening.gif

@davidmascharka this one is going to be a doozy. Hopefully we can review this in waves.

I am torn as to whether or not I want to add a mypy stage to the CI. The hacks that mypy has required in hypothesis (e.g. `if False` blocks  and no support for `Real`) are really gross.

@Zac-HD if you wouldn't mind occasionally taking a look at this PR, your experience with type hinting would be much appreciated!

### DIFF
--- a/src/mygrad/tensor_base.py
+++ b/src/mygrad/tensor_base.py
@@ -3,9 +3,20 @@ This module defines the base tensor class along with all of its essential
 attributes and special methods. Public math methods, e.g. ``sum``, ``mean``,
 etc., are bound to the Tensor class in ``mygrad.__init__.py``.
 """
-
 from functools import wraps
-from typing import Optional, Set, Type, Union
+from numbers import Real
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+)
 
 import numpy as np
 
@@ -25,8 +36,11 @@ from mygrad.operation_base import BroadcastableOp, Operation
 from mygrad.tensor_core_ops.indexing import GetItem, SetItem
 from mygrad.tensor_manip.array_shape.ops import Flatten
 from mygrad.tensor_manip.transpose_like.ops import Tensor_Transpose_Property
+from mygrad.typing import ArrayLike
 
 __all__ = ["Tensor"]
+
+T = TypeVar("T")
 
 
 class Tensor:
@@ -113,7 +127,13 @@ class Tensor:
     __array_priority__ = 15.0
 
     def __init__(
-        self, x, *, dtype=None, constant=False, _scalar_only=False, _creator=None
+        self,
+        x: ArrayLike,
+        *,
+        dtype: Optional[type] = None,
+        constant: bool = False,
+        _scalar_only: bool = False,
+        _creator: Optional[Operation] = None
     ):
         """
         Parameters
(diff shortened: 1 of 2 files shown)


src/mygrad/tensor_base.py
```diff
@@ -25,9 +36,12 @@
 from mygrad.tensor_core_ops.indexing import GetItem, SetItem
 from mygrad.tensor_manip.array_shape.ops import Flatten
 from mygrad.tensor_manip.transpose_like.ops import Tensor_Transpose_Property
+from mygrad.typing import ArrayLike
 
 __all__ = ["Tensor"]
 
+T = TypeVar("T")
```
Zac-HD: I'd define all your typevars in `mygrad.typing`; in hindsight we should have made a `hypothesis._typing` module.

src/mygrad/tensor_base.py
```diff
@@ -431,7 +451,7 @@ def null_gradients(self, clear_graph=True):
         """
         self._null_gradients(clear_graph=clear_graph, seen=set())
 
-    def _null_gradients(self, *, clear_graph: bool, seen: Set["Tensor"]):
+    def _null_gradients(self, *, clear_graph: bool, seen: Set["Tensor"]) -> None:
```
Zac-HD: A `Set[Tensor]` could be very large, right?  Not sure it helps here, but in general I'd be looking for places to use various [`weakref` collections](https://docs.python.org/3/library/weakref.html) to avoid keeping them around for longer than necessary.

src/mygrad/tensor_base.py
```diff
@@ -879,7 +901,7 @@ def T(self):
 
 
 # set all comparison operators - mirrors ndarray methods
-def tensor_to_array_wrapper(func):
+def tensor_to_array_wrapper(func: Callable) -> Callable:
```
Zac-HD: It's not always worth specifying argument types, but specifying the return types of these functions could be useful.

src/mygrad/typing.py
```diff
@@ -0,0 +1,6 @@
+from typing import Union, Sequence, Any
+from numbers import Real
+from numpy import ndarray
+
+
+ArrayLike = Union[Real, ndarray, Sequence[Any]]
```
Zac-HD:
```python
if typing.TYPE_CHECKING:

ArrayLike = Union[Real, ndarray, "Tensor", Sequence[Any]]
```

is inelegant, but would simplify *many* other type-annotations.  The import cycle only breaks things if it's executed at runtime, after all!

```python
# Sequence[Any] is imprecise; we actually only want equally-nested sequences that end in Real
s1 = Sequence[Real]
s2 = Sequence[s1]
...
s32 = Sequence[s31]
seq_arraylike = Union[s1, s2, ..., s32]  # Sequence[Union[s1, s2]] is *not* valid!

ArrayLike = Union[Real, ndarray, seq_arraylike]
```

is considerably more verbose still, but IMO not unreasonable: it's nicely isolated in your typing helpers module, and can catch a range of issues with element types and ragged-dimensional arrays.

src/mygrad/tensor_base.py
rsokl: Oh wow, I didn't even know about the `weakref` collections! This will be very very useful. Thank you! ❤️

Zac-HD: The standard library strikes again!  So many useful data structures and tools, so little advertising...

rsokl:
Closed by https://github.com/rsokl/MyGrad/pull/354

But still plenty of work to be done here.

Issue #102, opened 2018-11-25
Implement weighted average


rsokl:
See [numpy.average](https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.average.html#numpy.average)

bantmen:
Hey, I could take a stab at this if the issues are not stale. If all goes well, I could give https://github.com/rsokl/MyGrad/issues/117 a try too.

rsokl:
Please do! That would be great. Let me know if you have any questions about the development process, including code / doc styles and testing.

#117 is a really neat one too - it will be more involved, but feel free to take a swing at it.

bantmen:
Perfect. I'll put up a PR, then we can move on from there.

bantmen:
Would it be possible to backprop over the weights?

rsokl:
@bantmen sorry for missing this. I do think that we should permit backprop through the weights. Since they need not be normalized, it would be equivalent to backpropping through:

```python
(a * weights).sum(axis) / weights.sum(axis)
```

bantmen:
@rsokl No problem! We can continue the discussion at the PR.

rsokl:
Currently in progress at #215

PR #215, opened 2020-03-09
Add Average operation

bantmen:
**Background**
This is my attempt at implementing weighted average – see https://github.com/rsokl/MyGrad/issues/102.

**Example usage**
``` python3
>>> from mygrad import Tensor, average
>>> import numpy as np
>>> x = Tensor(np.array([[1, 10], [1, 10]]))
>>> weights = Tensor(np.array([[1, 1], [10, 1]]))
>>> avg, scl = average(x, axis=None, weights=weights, returned=True)
>>> type(avg), type(scl)
(<class 'mygrad.tensor_base.Tensor'>, <class 'mygrad.tensor_base.Tensor'>)
>>> avg.backward()
>>> x.grad
array([[0.07692308, 0.07692308],
       [0.76923077, 0.07692308]])
>>> weights.grad
array([[-0.10650888,  0.58579882],
       [-0.10650888,  0.58579882]])
```

**Rough edges**
- I don't like that `_canonicalize_weights` is exported from `sequential/funcs.py` – it's currently used for a hypothesis strategy + there are some tests targeting it.
- I feel like I have broken some patterns  – is it OK that `average` does not have an implementation in `sequential/ops.py`? I initially had an ops.py implementation, but I felt like the boilerplate did not add value.
- `average` documentation is missing: Should I copy over the numpy documentation and add some mygrad examples? 
- It would have been nice to have a linter command that I could run (which would apply black and isort on my files) – should I be running any autoformatters?
- The testing utility doesn't seem to support tuple outputs (as is the case when `returned=True`). The CI seems to have caught that I am not testing that branch (which is great).

### DIFF
--- a/src/mygrad/__init__.py
+++ b/src/mygrad/__init__.py
@@ -29,8 +29,6 @@ for attr in (
     cumprod,
     cumsum,
     mean,
-    _canonicalize_weights,
-    average,
     std,
     var,
     max,
--- a/src/mygrad/__init__.py
+++ b/src/mygrad/__init__.py
@@ -29,6 +29,8 @@ for attr in (
     cumprod,
     cumsum,
     mean,
+    _canonicalize_weights,
+    average,
     std,
     var,
     max,
--- a/src/mygrad/math/sequential/funcs.py
+++ b/src/mygrad/math/sequential/funcs.py
@@ -1,5 +1,6 @@
 import numpy as np
 
+import mygrad
 from mygrad.tensor_base import Tensor
 
 from .ops import *
@@ -167,16 +168,14 @@ def _canonicalize_weights(x, axis, weights):
         return weights
     if axis is None:
         raise TypeError(
-            "Axis must be specified when shapes of a and weights "
-            "differ.")
+            "Axis must be specified when shapes of a and weights " "differ."
+        )
     if weights.ndim != 1:
-        raise TypeError(
-            "1D weights expected when shapes of a and weights differ.")
+        raise TypeError("1D weights expected when shapes of a and weights differ.")
     if weights.shape[0] != x.shape[axis]:
-        raise ValueError(
-            "Length of weights not compatible with specified axis.")
+        raise ValueError("Length of weights not compatible with specified axis.")
     # setup weights to broadcast along axis
-    weights = np.broadcast_to(weights, (x.ndim-1)*(1,) + weights.shape)
+    weights = mygrad.broadcast_to(weights, (x.ndim - 1) * (1,) + weights.shape)
     return weights.swapaxes(-1, axis)
 
 
@@ -187,12 +186,11 @@ def average(x, axis=None, weights=None, constant=False, returned=False):
         weights = _canonicalize_weights(x, axis, weights)
         scl = sum(weights, axis=axis, constant=constant)
         if np.any(scl == 0.0):
-            raise ZeroDivisionError(
-                "Weights sum to zero, can't be normalized")
+            raise ZeroDivisionError("Weights sum to zero, can't be normalized")
         avg = sum(x * weights, axis=axis, constant=constant) / scl
     if returned:
         if avg.shape != scl.shape:
-            scl = np.broadcast_to(scl, avg.shape)
+            scl = mygrad.broadcast_to(scl, avg.shape)
         return avg, scl
     else:
         return avg
--- a/src/mygrad/math/sequential/funcs.py
+++ b/src/mygrad/math/sequential/funcs.py
@@ -168,7 +168,7 @@ def _canonicalize_weights(x, axis, weights):
         return weights
     if axis is None:
         raise TypeError(
-            "Axis must be specified when shapes of a and weights " "differ."
+            "Axis must be specified when shapes of a and weights differ."
         )
     if weights.ndim != 1:
         raise TypeError("1D weights expected when shapes of a and weights differ.")
@@ -181,13 +181,13 @@ def _canonicalize_weights(x, axis, weights):
 
 def average(x, axis=None, weights=None, constant=False, returned=False):
     if weights is None:
-        avg = mean(x, axis=axis, constant=constant)
-    else:
-        weights = _canonicalize_weights(x, axis, weights)
-        scl = sum(weights, axis=axis, constant=constant)
-        if np.any(scl == 0.0):
-            raise ZeroDivisionError("Weights sum to zero, can't be normalized")
-        avg = sum(x * weights, axis=axis, constant=constant) / scl
+        weights = Tensor(np.ones(x.shape, dtype=x.dtype))
+
+    weights = _canonicalize_weights(x, axis, weights)
+    scl = sum(weights, axis=axis, constant=constant)
+    if np.any(scl == 0.0):
+        raise ZeroDivisionError("Weights sum to zero, can't be normalized")
+    avg = sum(x * weights, axis=axis, constant=constant) / scl
     if returned:
         if avg.shape != scl.shape:
             scl = mygrad.broadcast_to(scl, avg.shape)
--- a/src/mygrad/math/sequential/funcs.py
+++ b/src/mygrad/math/sequential/funcs.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 from mygrad.tensor_base import Tensor
 
 from .ops import *
@@ -5,6 +7,8 @@ from .ops import *
 __all__ = [
     "sum",
     "mean",
+    "_canonicalize_weights",
+    "average",
     "var",
     "std",
     "amax",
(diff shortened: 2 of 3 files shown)


rsokl:
This is looking good! I plan to review this over the weekend :)

src/mygrad/__init__.py
```diff
@@ -29,6 +29,8 @@
     cumprod,
     cumsum,
     mean,
+    _canonicalize_weights,
```
rsokl: This should not be included in the top-level namespace of mygrad

```suggestion
```

src/mygrad/math/sequential/funcs.py
```diff
@@ -1,10 +1,14 @@
+import numpy as np
+
 from mygrad.tensor_base import Tensor
 
 from .ops import *
 
 __all__ = [
     "sum",
     "mean",
+    "_canonicalize_weights",
```
rsokl: We don't want `_canonicalize_weights` to be included if someone does `import *`

```suggestion
```

src/mygrad/math/sequential/funcs.py
```diff
@@ -157,6 +161,44 @@ def mean(x, axis=None, keepdims=False, constant=False):
...
+            "Axis must be specified when shapes of a and weights "
+            "differ.")
+    if weights.ndim != 1:
+        raise TypeError(
+            "1D weights expected when shapes of a and weights differ.")
+    if weights.shape[0] != x.shape[axis]:
+        raise ValueError(
+            "Length of weights not compatible with specified axis.")
+    # setup weights to broadcast along axis
+    weights = np.broadcast_to(weights, (x.ndim-1)*(1,) + weights.shape)
```
rsokl: `weights` can be a `Tensor` - you need to use [`mygrad.broadcast_to`](https://mygrad.readthedocs.io/en/latest/generated/mygrad.broadcast_to.html#mygrad.broadcast_to) here.

src/mygrad/math/sequential/funcs.py
```diff
@@ -157,6 +161,44 @@ def mean(x, axis=None, keepdims=False, constant=False):
...
+    else:
+        weights = _canonicalize_weights(x, axis, weights)
+        scl = sum(weights, axis=axis, constant=constant)
+        if np.any(scl == 0.0):
+            raise ZeroDivisionError(
+                "Weights sum to zero, can't be normalized")
+        avg = sum(x * weights, axis=axis, constant=constant) / scl
+    if returned:
+        if avg.shape != scl.shape:
+            scl = np.broadcast_to(scl, avg.shape)
```
rsokl: `scl` is a `Tensor` - you need to use [`mygrad.broadcast_to`](https://mygrad.readthedocs.io/en/latest/generated/mygrad.broadcast_to.html#mygrad.broadcast_to) here.

tests/math/sequence/test_sequential_funcs.py
```diff
@@ -46,6 +46,31 @@ def ddof_arg(*arrs):
...
+    # - the same shape as arr OR
+    # - 1D with compatible length
+    # See: https://github.com/numpy/numpy/blob/c31cc36a8a814ed4844a2a553454185601914a5a/numpy/lib/function_base.py#L397
+    wt_same_shape, wt_1D = range(2)
+    if arr.ndim == 0 or draw(st.one_of(st.just(wt_same_shape), st.just(wt_1D))) == wt_same_shape:
+        axis = draw(axis_arg(arr))
+        wt_shape = arr.shape
+    else: # wt_1D
+        # Only integer axis is supported for 1D weights
+        axis = draw(st.integers(min_value=0, max_value=len(arr.shape) - 1))
```
rsokl: Negative axes are permitted:

```suggestion
        axis = draw(st.integers(min_value=-arr.ndim, max_value=arr.ndim - 1))
```

tests/math/sequence/test_sequential_funcs.py
```diff
@@ -46,6 +46,31 @@ def ddof_arg(*arrs):
...
+        wt_shape = (arr.shape[axis],)
+    # Filter any axis summing up to 0 to avoid ZeroDivisionError
+    wt_strat = hnp.arrays(
+        dtype=np.float,
+        shape=wt_shape,
+        elements=st.floats(allow_infinity=False, allow_nan=False, min_value=-10, max_value=10),
```
rsokl: Specifying a min and max in `floats` will automatically preclude infs/nans from being generated

```suggestion
        elements=st.floats(min_value=-10, max_value=10),
```

tests/math/sequence/test_sequential_funcs.py
```diff
@@ -46,6 +46,31 @@ def ddof_arg(*arrs):
...
+    ).filter(
+        lambda wt: not np.isclose( _canonicalize_weights(arr, axis, wt).sum(axis=axis), 0).any()
+    )
+    weights = draw(wt_strat | st.none())
```
rsokl: It is preferable to shrink towards simpler strategies

```suggestion
    weights = draw(st.none() | wt_strat)
```

tests/math/sequence/test_sequential_funcs.py
```diff
@@ -8,7 +8,7 @@
 from pytest import raises
 
 import mygrad
-from mygrad import amax, amin, cumprod, cumsum, mean, prod, std, sum, var
+from mygrad import amax, amin, cumprod, cumsum, mean, _canonicalize_weights, average, prod, std, sum, var
```
rsokl: Move the import of `_canonicalize_weights` to be directly from `mygrad.math.sequential.funcs`

src/mygrad/__init__.py
```diff
@@ -29,6 +29,8 @@
...
+    average,
```
rsokl: `average` is not a method of `numpy.ndarray` so we ought not include it as a method on `Tensor`

```suggestion
```

tests/math/sequence/test_sequential_funcs.py
bantmen: Interesting, where can I read more about that? I found https://hypothesis.readthedocs.io/en/latest/data.html#shrinking

bantmen:
Re the granular "suggestion" commits, we can squash commit.

tests/math/sequence/test_sequential_funcs.py
rsokl: [`st.one_of`](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.one_of)

> Examples from this strategy will generally shrink to ones that come from strategies earlier in the list, then shrink according to behaviour of the strategy that produced them. In order to get good shrinking behaviour, try to put simpler strategies first. e.g. `one_of(none(), text())` is better than `one_of(text(), none())`.

bantmen:
@rsokl I've addressed your comments and squashed the granular commits to reduce noise.

tests/math/sequence/test_sequential_funcs.py
```diff
@@ -357,7 +437,6 @@ def test_prod_bkwd():
 )
 def test_multi_zero_prod_bkwd():
     """Drives tests cases with various configurations of zeros"""
-    pass
```
bantmen: Not sure how I feel about this but `tox` prefers it this way.

tests/math/sequence/test_sequential_funcs.py
```diff
@@ -183,6 +183,18 @@ def test_average_weights_zero_sum():
...
+def test_average_scl():
+    x = mg.Tensor(np.array([1, 2, 3, 4]))
+    weights = mg.Tensor(np.array([1, 2, 3, 4]))
+    avg, scl = average(x, weights=weights, returned=True)
+    expected_avg, expected_scl = np.average(x.data, weights=weights.data, returned=True)
+    assert np.isclose(avg.data, expected_avg)
+    assert np.isclose(scl.data, expected_scl)
+    scl.backward()
+    assert avg.grad is None
+    assert weights.grad is not None
```
rsokl: You should add a check that the gradient for the weights is indeed correct.

tests/math/sequence/test_sequential_funcs.py
rsokl: `tox`? Do you mean the linter that gets run when you run the `format` job?

bantmen: I was referring to `tox -e format` from https://github.com/rsokl/MyGrad/blob/master/CONTRIBUTING.md.

tests/math/sequence/test_sequential_funcs.py
bantmen: Good call, like you pointed out a branch of `average` was completely untested where `scl` was being read even though it was not defined.

tests/math/sequence/test_sequential_funcs.py
```diff
@@ -141,6 +171,81 @@ def test_mean_bkwd():
...
+
+@backprop_test_factory(
+    mygrad_func=average_returned_wrapper(average),
+    true_func=average_returned_wrapper(np.average),
+    num_arrays=1,
+    kwargs=gen_average_args,
+    index_to_bnds={0: (-10, 10)},
+    vary_each_element=True,
+)
+def test_average_scl_bkwd():
```
bantmen: @rsokl I have added a test for `scl` which is awkward because gradient of `scl` with respect to the input is 0. So the numerical derivative is returning 0 while `x.grad` is returning `None`, after `scl.backward()` is run, because `x` is not part of `scl`'s computational graph. 

If we want, one way to handle this is to `assert_allclose` between `np.broadcast_to(0, d_num.shape)` and `d_num` when `arr.grad` is `None` at https://github.com/bantmen/MyGrad/blob/c6d6d878596aba65e98ea129d280485e67111c01/tests/wrappers/uber.py#L632

bantmen:
If I understand correctly, the tests wrapped with `backprop_test_factory` are currently testing the gradient with respect to the `x` input, what would be the conventional way to test the gradient with respect to the other input i.e. `weights`?

tests/math/sequence/test_sequential_funcs.py
```diff
@@ -141,6 +171,81 @@ def test_mean_bkwd():
...
+@fwdprop_test_factory(
+    mygrad_func=average, true_func=np.average, num_arrays=1, kwargs=gen_average_args,
+)
+def test_average_fwd():
+    pass
+
+
+def test_average_weights_zero_sum():
+    with raises(ZeroDivisionError):
+        average(np.array([1, 2]), weights=np.array([-1, 1]))
```
rsokl, line 183:
```suggestion
        average(np.array([0, 2]), weights=np.array([-1, 1]))
```

rsokl:
@bantmen hmm this is a frustrating limitation of `backprop_test_factory` as it stands. Your "hack" using `average_returned_wrapper` is in the right direction, but I think I just need to make an improvement to `backprop_test_factory`  to make things easier

rsokl:
Given the update in #222 you can now indicate that `weights` should also be treated as a variable in the computational graph. 

Modify your usage of `backprop_test_factory ` by specifying:
    - `num_arrays=1`
    - `arrs_from_kwargs={1: "weights"}`

And modify your wrapper slightly, interchanging the ordering of `axis` and `weights` in the lambda definition:

```python
def average_returned_wrapper(average_func):
    return lambda x, weights, axis: average_func(x, axis=axis, weights=weights, returned=True)[1]
```

This should exercise the gradient in `weights`. Still a little bit hacky, but I'll have to do a more thorough change to the testing suite to make this more natural.

rsokl:
Closing this for now. Will resurrect when MyGrad 2.0 comes out and I have some time

Issue #210, opened 2020-02-18
Iteration over 0D tensor should raise type error


rsokl:
Found in #209 

Calling `sum` (std-lib) on a 0D numpy array raises a type error whereas calling it on a MyGrad tensor simply returns 0 (the tensor acts like an empty iterator). `Tensor` should be updated to match numpy's behavior.

PR #216, opened 2020-03-10
Iteration over 0D tensor should raise type error

aslvrstn:
Fixes https://github.com/rsokl/MyGrad/issues/210

This is technically a breaking change I think, in that theoretically someone could be depending on it. Up to you how much to worry about that.

### DIFF
--- a/src/mygrad/tensor_base.py
+++ b/src/mygrad/tensor_base.py
@@ -581,6 +581,10 @@ class Tensor:
         return self.data.__contains__(item)
 
     def __getitem__(self, item):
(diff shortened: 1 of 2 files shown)


rsokl:
@aslvrstn this looks good! Do you think it is worth it to test that numpy arrays have the same behavior? I'm thinking it is probably too paranoid... Anyway, feel free to merge!

aslvrstn:
I thought about trying to express that, but I couldn't come up with a good way. Like, I could test that numpy raises when you do the same thing, but then it's just a test of numpy? FWIW I confirmed that it does raise in the same way. I feel like the comment in the code is enough to express why.

PR #221, opened 2020-04-15
Distinguish backprop tests for Hypothesis example DB

Zac-HD:
As discussed, this PR illustrates how to fix collisions from test factories in the Hypothesis example database... without actually fixing them.  But you can take it from here and I'll review it at the end :grin:

### DIFF
--- a/tests/wrappers/uber.py
+++ b/tests/wrappers/uber.py
@@ -27,6 +27,45 @@ def _to_dict(x):
     return x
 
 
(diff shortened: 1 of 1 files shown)


rsokl:
You set a dangerous precedent here, @Zac-HD 

You: "I wont actually fix this"
You, a couple of months later: "Here's a fix!"

:)

Zac-HD:
Yeah, but now I only have... uh... ten open pull requests!  That's progress, right?

PR #223, opened 2020-04-18
Drop py35

rsokl:
🥳


tests/test_utils.py
```diff
@@ -13,6 +13,11 @@
 from tests.custom_strategies import broadcastable_shapes, everything_except
 
 
+@st.composite
+def f(draw):
+    return 1
+
+
```
davidmascharka, line 20: was this supposed to be here?

rsokl: ha nooooope!

PR #226, opened 2020-06-20
Add leaky_relu activation function

davidmascharka:
Trying to get back to merging MyNN :smile: 

Strategically, do we want to merge things into master like this (where possible) or should this all just use mygrad-2.0 as the base and then merge everything at once?


rsokl:
>Strategically, do we want to merge things into master like this (where possible) or should this all just use mygrad-2.0 as the base and then merge everything at once?

I think it is fine to add things piece-by-piece where possible. It certainly makes it easier to review. It also means that we can do incremental releases without having to worry about "well this isn't _really_ 2.0 yet"

PR #230, opened 2020-06-20
Hotfix leaky relu test

davidmascharka:
What's the right way to fix this? The issue is that the actual slope value is ok but the gradient being passed back through this operation is randomly in (-10, 10) which can cause an inf


rsokl:
Yeah I think this is sufficient. If this keeps rearing its head then it is probably worth it to just throw conservative bounds on these parameters

PR #231, opened 2020-06-21
Add GLU activation function

davidmascharka:
Some craziness in the test because of the conditions this thing requires :grimacing: 

Last one! :partying_face: :tada:

### DIFF
--- a/src/mygrad/nnet/activations/glu.py
+++ b/src/mygrad/nnet/activations/glu.py
@@ -1,0 +1,53 @@
+from mygrad import multiply
+from .sigmoid import sigmoid
+
+
+def glu(x, dim=-1, constant=False):
+    """ Returns the Gated Linear Unit A * σ(B), where A and B are split from `x`.
+
+    Parameters
+    ----------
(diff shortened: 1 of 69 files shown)


src/mygrad/nnet/activations/glu.py
```diff
@@ -0,0 +1,72 @@
+from numpy import ndarray
+
+from mygrad import multiply, Tensor
+from .sigmoid import sigmoid
+
+
+def glu(x, dim=-1, constant=False):
```
rsokl, line 7: I think this should be `axis` to stick with the parlance of NumPy

davidmascharka: :flushed: why did I do that? :facepalm: updated

davidmascharka:
@rsokl I didn't see a changelog - are we tracking changes? Was planning to bump to 1.6 and highlight that all these activations have been added

rsokl:
@davidmascharka I am planning to generate the `.rst` docs for these, then I'll bump to 1.6

PR #236, opened 2020-06-23
Add constant initializer

davidmascharka:
This feels a little silly....but I think it makes sense to put this here. We're going to end up adding things that are very much `nnet`-specific (he_uniform, glorot_normal, dirac, ...) so having this package makes sense. On a practical front, I think we could either:
- merge this
- change this to just alias `full`

The decision really rests on whether we want to do `*shape` or `shape` - I'm very slightly in favor of `*shape` because it'll mirror the interface for layers that we pull in (so we can do `conv(3, 3)` instead of `conv((3, 3))`) but if I hear an argument I'll drop it and go with plain `shape` :shrug:

### DIFF
--- a/tests/nnet/initializers/test_constant.py
+++ b/tests/nnet/initializers/test_constant.py
@@ -19,7 +19,9 @@ def test_constant_initializer(array, constant):
     value = array.reshape(-1)[0]  # let hypothesis do all the heavy lifting picking an acceptable value for the dtype
     assume(not np.isnan(value))
     tensor = constant_initializer(array.shape, value=value, dtype=array.dtype, constant=constant)
+
     assert isinstance(tensor, Tensor)
     assert np.allclose(tensor.data, value, equal_nan=True)
     assert tensor.data.shape == array.shape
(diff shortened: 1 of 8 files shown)


tests/nnet/initializers/test_constant.py
```diff
@@ -0,0 +1,22 @@
...
+        shape=hnp.array_shapes(),
+    ),
+    constant=st.booleans(),
+)
+def test_constant_initializer(array, constant):
+    value = array.reshape(-1)[0]  # let hypothesis do all the heavy lifting picking an acceptable value for the dtype
+    assume(not np.isnan(value))
+    tensor = constant_initializer(array.shape, value=value, dtype=array.dtype, constant=constant)
+    assert np.allclose(tensor.data, value, equal_nan=True)
+    assert np.all(tensor[:, np.newaxis] == tensor[np.newaxis])
```
rsokl, line 22: What is the purpose of this check?

davidmascharka: yeah that was a weird holdover. removed and added some extra validation on other attributes

PR #242, opened 2020-06-23
Add dirac initializer

davidmascharka:
Pushing this here to get some discussion on the tests. When you convolve a dirac-initialized filter with a tensor you should get back out (roughly) that tensor. I tried doing something like this:

``` python
@given(
    data=st.data(),
    shape=hnp.array_shapes(min_dims=1, max_dims=3, max_side=10),
    num_filters=st.integers(1, 10),
    num_batch=st.integers(1, 10),
    num_channel=st.integers(1, 10),
)
def test_dirac_initializer(data, shape, num_filters, num_batch, num_channel):
    img_shape = (num_batch, num_channel) + shape
    win_dim = len(shape)
    shape = (num_batch, num_channel) + shape
    win_shape = data.draw(
        st.tuples(*(st.integers(1, s) for s in shape[-win_dim:])), label="win_shape"
    )
    kernel_shape = (num_filters, shape[1], *win_shape)
    x = data.draw(
        hnp.arrays(dtype=float, shape=img_shape, elements=st.floats(-10, 10)), label="x"
    )
    kernel = dirac(kernel_shape)

    out = conv_nd(x, kernel, stride=1, padding=1)
    assert isinstance(out, Tensor)
    x = x[:, :min(num_filters, num_channel)]

    assert np.allclose(out.data, x)
```

(mostly stolen from the conv test) but it's difficult to figure out what the shape should be. Hoping you can provide a suggestion on that.


davidmascharka:
Example of a nicely-behaved case:

``` python
>>> kernel = dirac(1, 3, 3, 3)
>>> x = np.random.rand(3, 1, 10, 10)
>>> out = conv_nd(x, kernel, stride=1, padding=1)
>>> np.allclose(x, out.data)
True
```

but `array_shapes()` is giving me things that are harder to deal with like:

``` python
>>> kernel = dirac(1, 2, 1); kernel
Tensor([[[1.],
         [0.]]], dtype=float32)

>>> x = np.random.rand(1, 2, 1); x
array([[[0.15250423],
        [0.04218379]]])

>>> conv_nd(x, kernel, stride=1, padding=1)
Tensor([[[0.        , 0.15250423, 0.        ]]])
```

where I basically need to remove the padding and then check that the output is a subset of the input

rsokl:
I'll have to loop back to this when I have a little bit more time to dig into it

rsokl:
(running against new CI)

rsokl:
haha nothing to see here, just merging this PR over 9 months since it opened :P

PR #248, opened 2020-06-25
Add he_normal activation

davidmascharka:
Bad branching so you can ignore the changes to test_glorot_normal :/ those are in a different pr and I'll make sure that merges first

### DIFF
--- a/src/mygrad/nnet/initializers/he_normal.py
+++ b/src/mygrad/nnet/initializers/he_normal.py
@@ -42,7 +42,7 @@ def he_normal(*shape, gain=1, dtype=np.float32, constant=False):
     using the `gain` variable above.
     """
     if not np.issubdtype(dtype, np.floating):
-        raise ValueError("Glorot Normal initialization requires a floating-point dtype")
+        raise ValueError("He Normal initialization requires a floating-point dtype")
 
     if len(shape) == 1:
         shape = shape[0]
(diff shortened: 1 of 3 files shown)


src/mygrad/nnet/initializers/he_normal.py
```diff
@@ -0,0 +1,53 @@
...
+    if not np.issubdtype(dtype, np.floating):
+        raise ValueError("He Normal initialization requires a floating-point dtype")
+
+    if len(shape) == 1:
+        shape = shape[0]
+    if len(shape) < 2:
+        raise ValueError("He Normal initialization requires at least two dimensions")
+
+    std = gain / np.sqrt(shape[1] * (np.prod(shape[2:]) if len(shape) > 2 else 1))
+    return Tensor(np.random.normal(0, std, shape), dtype=dtype, constant=constant)
```
rsokl, line 53: Similar to He uniform, it seems like this should just call into `mygrad.normal` - then we don't have to duplicate the validation logic / tests

src/mygrad/nnet/initializers/he_normal.py
```diff
@@ -0,0 +1,53 @@
...
+    https://arxiv.org/abs/1502.01852
+
+    A Tensor :math:`W` initialized in this way should be drawn from a distribution about
+
+    .. math::
+        \mathcal{N}(0, \sqrt{\frac{2}{(1+a^2)n_l}})
+
+    where :math:`a` is the slope of the rectifier following this layer, which is incorporated
+    using the `gain` variable above.
+    """
```
rsokl, line 43: Examples section

PR #249, opened 2020-06-25
Add he_uniform initializer

davidmascharka:


src/mygrad/nnet/initializers/he_uniform.py
```diff
@@ -0,0 +1,56 @@
...
+    if len(shape) == 1:
+        shape = shape[0]
+    if len(shape) < 2:
+        raise ValueError("He Uniform initialization requires at least two dimensions")
+
+    if isinstance(gain, Tensor):
+        gain = gain.item()
+
+    bound = gain / np.sqrt(3 / shape[1] * (np.prod(shape[2:]) if len(shape) > 2 else 1))
+    return Tensor(np.random.uniform(-bound, bound, shape), dtype=dtype, constant=constant)
```
rsokl, line 56: Can't this all be simplified by just calling into `mygrad.uniform` with the appropriate bounds?

src/mygrad/nnet/initializers/he_uniform.py
```diff
@@ -0,0 +1,56 @@
...
+    constant : bool, optional (default=False)
+        If `True`, the returned tensor is a constant (it
+            does not back-propagate a gradient).
+
+    Returns
+    -------
+    mygrad.Tensor, shape=``shape``
+        A Tensor, with values initialized according to the He uniform initialization.
+
+    Extended Description
```
rsokl, line 29: This is great. I would really like to go through and add "takeaways" to these sections. I.e. "this paper recommends using this initialization for X circumstances; it makes such and such assumptions"

These are things that I end up having to re-read the paper for and wish they were just documented. This need not be done for this round of PRs, but it is something for us to keep in mind. I also think that will help make apparent the value of `mygrad` as an educational tool :)

src/mygrad/nnet/initializers/he_uniform.py
```diff
@@ -0,0 +1,56 @@
...
+    https://arxiv.org/abs/1502.01852
+
+    A Tensor :math:`W` initialized in this way should be drawn from a distribution about
+
+    .. math::
+        U[-\sqrt{\frac{6}{(1+a^2)n_l}}, \sqrt{\frac{6}{(1+a^2)n_l}}]
+
+    where :math:`a` is the slope of the rectifier following this layer, which is incorporated
+    using the `gain` variable above.
+    """
```
rsokl, line 43: Examples section

src/mygrad/nnet/initializers/he_uniform.py
davidmascharka: That's a great addition; definitely more informative than just having the paper linked

PR #252, opened 2020-06-25
Add normal initializer

davidmascharka:
Last one :sweat_smile: 

Do we want to put `normal` and `uniform` into `tensor_creation` as well? Are we going to mirror `np.random` in the future?

### DIFF
--- a/src/mygrad/nnet/initializers/normal.py
+++ b/src/mygrad/nnet/initializers/normal.py
@@ -1,0 +1,43 @@
+import numpy as np
+from mygrad import Tensor
+
+def normal(*shape, mean=0, std=1, dtype=np.float32, constant=False):
+    """ Initialize a :class:`mygrad.Tensor` by drawing from a normal (Gaussian) distribution.
+
+    Parameters
+    ----------
+    shape : Sequence[int]
+        The output shape.
+
+    mean : Real, optional (default=0)
+        The mean of the distribution from which to draw.
+
+    std : Real, optional (default=1)
+        The standard deviation of the distribution from which to draw.
+
+    dtype : data-type, optional (default=float32)
+        The data type of the output tensor; must be a floating-point type.
+
+    constant : bool, optional (default=False)
+        If `True`, the returned tensor is a constant (it
+            does not back-propagate a gradient).
+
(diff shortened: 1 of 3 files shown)


src/mygrad/nnet/initializers/normal.py
```diff
@@ -0,0 +1,43 @@
...
+        If `True`, the returned tensor is a constant (it
+            does not back-propagate a gradient).
+
+    Returns
+    -------
+    mygrad.Tensor, shape=`shape`
+        A Tensor, with values drawn from Ɲ(μ, σ²), where μ=`mean` and σ=`std`.
+    """
+    if not np.issubdtype(dtype, np.floating):
+        raise ValueError("Glorot Normal initialization requires a floating-point dtype")
```
rsokl, line 31:
```suggestion
        raise ValueError("Normal initialization requires a floating-point dtype")
```

src/mygrad/nnet/initializers/normal.py
```diff
@@ -0,0 +1,43 @@
...
+
+    constant : bool, optional (default=False)
+        If `True`, the returned tensor is a constant (it
+            does not back-propagate a gradient).
+
+    Returns
+    -------
+    mygrad.Tensor, shape=`shape`
+        A Tensor, with values drawn from Ɲ(μ, σ²), where μ=`mean` and σ=`std`.
+    """
```
rsokl, line 29: I know this is pesky, but I do think we should strive to have an Examples section in each docstring - especially since students will be encountering this

rsokl:
>Do we want to put normal and uniform into tensor_creation as well? Are we going to mirror np.random in the future?

Let's hold off for now. I do think we might mirror `numpy.random` down the road, but I don't anticipate doing that in the near future. I would be really interested in taking a crack at doing a tensorflow probability sort of thing there... but that might be very hard

src/mygrad/nnet/initializers/normal.py
davidmascharka: :sweat_smile: yeah I realized all of these were missing examples as I was going to bed :smile: will update them all

PR #253, opened 2020-06-26
Add negative log-likelihood loss

davidmascharka:


src/mygrad/nnet/losses/negative_log_likelihood.py
```diff
@@ -0,0 +1,65 @@
+import numpy as np
+
+from mygrad import Tensor, sum
```
rsokl, line 3: Nit: let's import sum as `mygrad_sum` or something like that. Just so we can't unwittingly delete this import and still have the code "work"... but super slowly

src/mygrad/nnet/losses/negative_log_likelihood.py
```diff
@@ -0,0 +1,65 @@
...
+    Parameters
+    ----------
+    x : array_like, shape=(N, C)
+        The C log-probabilities for each of the N pieces of data.
+
+    y_true : array_like, shape=(N,)
+        The correct class indices, in [0, C), for each datum.
+
+    weights : array_like, optional (default=None)
+        The weighting factor to use on each class, or None.
```
rsokl, line 22: This should be documented as shape-(C,), right?

src/mygrad/nnet/losses/negative_log_likelihood.py
```diff
@@ -0,0 +1,65 @@
...
+    >>> negative_log_likelihood(x, y_true)
+    Tensor(4.6)
+    """
+    if isinstance(y_true, Tensor):
+        y_true = y_true.data
+
+    check_loss_inputs(x, y_true)
+
+    label_locs = (range(len(y_true)), y_true)
+    factors = weights[y_true] if weights is not None else np.ones_like(y_true)
```
rsokl, line 64: Do you think we should do an input validation on the shape of `weights`: ndim-1, shape-(C,)?

rsokl:
@davidmascharka I see these "marked as resolved" but I do not see the changes - just in case you thought you pushed them.

davidmascharka:
heh the classic commit and don't push :flushed:

PR #254, opened 2020-06-26
Add focal loss

davidmascharka:


davidmascharka:
Oooh good catch. What do you think about switching this to

``` python
    label_locs = (range(len(targets)), targets)
    pc = scores[label_locs]
    raw_loss = sum(alpha * (1 - pc + 1e-14) ** gamma * log(pc))
    normalized_loss = raw_loss / (label_locs > 0).sum()
    return -mean(normalized_loss, constant=constant)
```

and noting that we assume class label 0 is background? I worry that if we make the user do anything additional that will be easily lost

rsokl:
Because of the potential obfuscation that can be done here (see: all of our previous work with focal loss 😓 ) I think we should return the loss on a per-datum level, and force the op to be scalar-only so that the user _must_ make a decision about aggregation. I.e. I think they should be forced to pick how to sum/normalize the loss. And we merely document the paper's crucial recommendation.

davidmascharka:
That's a fair point. How do you feel about returning the raw tensor for all the losses then and just having the user do whatever they want with it? In my mind, the pro is that it makes the interface to all the losses the same and it's as easy as calling `.mean()` or `.sum()` on the result. I think remembering to have a `.mean()` on all the losses is lower overhead than remembering that for all the losses except two you use them directly but one loss you have to reduce yourself. Our generic training loop would just change one line:

``` python
def train_epoch(train_data, train_labels, model, optim, batch_size=64):
    idxs = np.arange(len(train_data))
    np.random.shuffle(idxs)

    for batch in range(0, len(idxs), batch_size):
        batch_data = train_data[idxs[batch:batch+batch_size]]
        batch_labels = train_labels[idxs[batch:batch+batch_size]]

        outs = model(batch_data)
        loss = softmax_cross_entropy(outs, batch_labels)
        loss.mean().backward()  # instead of loss.backward()
        optim.step()
        loss.null_gradients()
```

which seems pretty minimal

rsokl:
> How do you feel about returning the raw tensor for all the losses then and just having the user do whatever they want with it?

This is exactly what I have in mind. No need to code up the various aggregations.

src/mygrad/nnet/losses/focal_loss.py
```diff
@@ -0,0 +1,151 @@
...
+        numpy.ndarray
+            The per-datum focal loss.
+        """
+        if isinstance(targets, Tensor):
+            targets = targets.data
+
+        check_loss_inputs(scores, targets)
+
+        self.variables = (scores,)
+        scores = np.copy(scores.data)
```
rsokl, line 59:
```suggestion
        scores = np.copy(scores.data if isinstance(scores, Tensor) else scores)
```

tests/nnet/losses/test_focal_loss.py
```diff
@@ -0,0 +1,100 @@
...
+@given(
+    num_datum=st.integers(1, 100),
+    num_classes=st.integers(1, 15),
+    alpha=st.floats(-1, 1),
+    gamma=st.floats(0, 5),
+    data=st.data(),
+    grad=st.floats(-1, 1),
+    target_type=st.sampled_from((np.array, Tensor)),
+)
+def test_softmax_focal_loss(num_datum, num_classes, alpha, gamma, data, grad, target_type):
```
rsokl, line 37: I think I will want to loop back to this and write tests using uber, since that makes nice extra checks; like protects against input mutation, propagation of `constant` etc. Not necessary for this PR but I'll make an issue about this.

PR #255, opened 2020-06-29
[MyGrad-2.0] Refactor losses to return per-datum

davidmascharka:


davidmascharka:
Oof I'll make a PR to bring 2.0 up to master first to minimize this diff

Issue #218, opened 2020-03-21
Add mygrad.any and Tensor.any


rsokl:
Should simply call `numpy.any` on the tensor's underlying numpy array.

samaocarpenter:
I'm gonna have a go at this

petarmhg:
Closed by #260.

PR #260, opened 2020-07-15
Add `mygrad.any` and `Tensor.any`

samaocarpenter:
Addresses #218

### DIFF
--- a/src/mygrad/math/nondifferentiable.py
+++ b/src/mygrad/math/nondifferentiable.py
@@ -45,3 +45,54 @@ def argmin(a, axis=None, out=None):
 
     a = a.data if isinstance(a, Tensor) else a
     return np.argmin(a, axis, out)
+
+def any(a, axis=None, out=None, keepdims=False):
+    """ Test whether any array or Tensor element along a given axis evaluates to True.
+
+    Returns single boolean if `axis` is ``None``
+
+    This documentation was adapted from ``numpy.add``
+
+    Parameters
+    ----------
+    a : array_like
+        
+    axis : None or int or tuple of ints, optional
+        Axis or axes along which a logical OR reduction is performed.
+        The default (``axis=None``) is to perform a logical OR over all
+        the dimensions of the input array. `axis` may be negative, in
+        which case it counts from the last to the first axis.
+        If this is a tuple of ints, a reduction is performed on multiple
+        axes, instead of a single axis or all the axes as before.
+
+    out : ndarray, optional
+        Alternate output array in which to place the result.  It must have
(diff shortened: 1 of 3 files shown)


src/mygrad/math/nondifferentiable.py
```diff
@@ -45,3 +45,55 @@ def argmin(a, axis=None, out=None):
...
+        in which case a reference to `out` is returned.
+
+    See Also
+    --------
+    Tensor.any : equivalent method
+
+    """
+
+    a = a.data if isinstance(a, Tensor) else a
+    return np.any(a, axis, out, keepdims)
```
rsokl, line 99: I recommend specifying the trailing arguments as keyword-arguments here.

tests/math/unary/test_nondifferentiable.py
```diff
@@ -51,3 +51,23 @@ def test_argmax(a, data):
...
+
+
+@given(
+    a=hnp.arrays(
+        shape=hnp.array_shapes(max_side=4, max_dims=5), dtype=dtype_strat_numpy
+    ),
+    data=st.data(),
+)
+def test_any(a, data):
+    axis = data.draw(valid_axes(ndim=a.ndim, single_axis_only=True), label="axis")
```
rsokl, line 63: Why do you have `single_axis_only=True` here? We want to also test multiple axes

samaocarpenter: Honestly, I have almost no idea how to use Hypothesis...I just consulted a couple tutorials and mostly copy-pasted the tests from above. I'm working on learning it so I can write better tests in the future - thank you for pointing this out!

samaocarpenter:
@rsokl I made those changes, thank you!

tests/math/unary/test_nondifferentiable.py
```diff
@@ -51,3 +51,23 @@ def test_argmax(a, data):
...
+    tensor = Tensor(a)
+
+    # tensor input
+    assert_allclose(mg.any(tensor, axis=axis), np.any(a, axis=axis))
+
+    # tensor method
+    assert_allclose(tensor.any(axis=axis), a.any(axis=axis))
+
+    # array input
+    assert_allclose(mg.any(a, axis=axis), np.any(a, axis=axis))
```
rsokl, line 73: `keepdims` also needs to be exercised in these tests - this parameter should also be drawn by Hypothesis.

samaocarpenter: This should be resolved - thanks!

rsokl:
Nice work @samaocarpenter ! Thanks for the contribution!

PR #263, opened 2020-07-22
float32 and float64 compatability in GRU

### LINKED ISSUE #261: The GRU has dtype compatibility issues
The GRU dies when dealing with mixes of float-32 and float64 data/parameters

petarmhg:
Closes #261 

Casts all parameters to the biggest dtype for any forward- or backward-pass calls to numba, and recasts to original data types as necessary when passing back gradients. Also updates GRU tests to check for both float32 and float64s.

The tests currently only have the data `X` as float64, as I am unsure of how to draw an array in the `given` decorator that has the `dtype` argument and `width` argument of `st.floats` agree

### DIFF
--- a/src/mygrad/nnet/layers/gru.py
+++ b/src/mygrad/nnet/layers/gru.py
@@ -14,8 +14,7 @@ except ImportError:  # pragma: no cover
 
 
 @vectorize(
-    ["int32(int32)", "int64(int64)", "float32(float32)", "float64(float64)"],
-    nopython=True,
+    ["float32(float32)", "float64(float64)"], nopython=True,
 )
 def sig(f):  # pragma: no cover
     """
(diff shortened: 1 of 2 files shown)


src/mygrad/nnet/layers/gru.py
```diff
@@ -203,22 +197,25 @@ def __call__(
             self.Wh,
             self.bh,
         )
+
+        self.type = max(t.dtype for t in self.variables)
```
rsokl, line 201: ooooh I was wondering how to do this

tests/nnet/layers/test_gru.py
```diff
@@ -265,22 +279,35 @@ def test_gru_backward(
     b_constants: Tuple[bool, bool, bool],
     X_constant: bool,
     V_constant: bool,
+    dtypes: Tuple[int, int, int, int],
```
petarmhg, line 282:
```suggestion
    dtypes: Tuple[np.dtype, np.dtype, np.dtype, np.dtype],
```

PR #265, opened 2020-07-27
use augmented add instead of add.at where possible in getitem

rsokl:
See https://github.com/numpy/numpy/issues/5922

Leads to a ~4x speedup in some initial cases:

image


rsokl:
This sounds good. It would be a nice optimization for sure!

PR #266, opened 2020-07-28
dispatch special cases for x ** 1 and x ** 2

rsokl:


rsokl:
True. Although those aren't much more efficient than the pow implementation

PR #267, opened 2020-07-28
make Tensor.constant settable

### LINKED ISSUE #92: Can't set `constant`
I'd like to be able to set tensors to constant, but I cannot:

``` python
from mygrad import Tensor
weights = Tensor([1])
weights.constant = True

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-43-58d8e7751d9c> in <module>
----> 1 weights.constant = True

AttributeError: can't set attribute
```

A workaround is to set `weights._constant`; is this behavior intentional?

rsokl:
Closes #92

### DIFF
--- a/tests/state_testing/test_state.py
+++ b/tests/state_testing/test_state.py
@@ -31,7 +31,7 @@ def _node_ID_str(num):
     return "v{}".format(num + 1)
 
 
-@settings(max_examples=200, deadline=None)
+@settings(max_examples=250, deadline=None)
 class GraphCompare(RuleBasedStateMachine):
     def __init__(self):
         super().__init__()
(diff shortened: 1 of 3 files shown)


tests/state_testing/test_state.py
```diff
@@ -31,7 +31,7 @@ def _node_ID_str(num):
     return "v{}".format(num + 1)
 
 
-@settings(max_examples=200, deadline=None)
+@settings(max_examples=250, deadline=None)
```
davidmascharka, line 34: why?

rsokl: Because I added a new rule, so I want the state machine to be able to explore similar depths of states despite that there is now more breadth to consider

PR #268, opened 2020-07-29
Improve softmax

rsokl:


src/mygrad/nnet/activations/softmax.py
```diff
@@ -9,48 +9,48 @@ def _softmax(x, kwargs):
...
     if np.issubdtype(x.dtype, np.integer):
         x = x.astype(np.float)
-    np.exp(x, out=x)
-    x /= x.sum(**kwargs)
+    if x.ndim > 0:
+        np.exp(x, out=x)
+        x /= x.sum(**kwargs)
+    else:
+        x = np.exp(x)
+        x = x / x.sum(**kwargs)
```
davidmascharka, line 17: isn't this branch just for scalars? In which case the output of softmax would just be 1 by definition so we don't need to do this computation?

src/mygrad/nnet/activations/softmax.py
```diff
@@ -85,54 +85,50 @@ def softmax(x, constant=False):
...
 
     Parameters
     ----------
-    x : array_like, shape=(D,) or shape=(N,D)
-        log-softmax is computed along the rows of ``x`` if
-        ``x`` is a 2D array. Otherwise log-softmax is computed
-        on the 1D ``x``.
+    x : array_like
+
+    axis : Union[None, int, Tuple[int, ...]], optional (default=-1)
```
davidmascharka, line 120: for these can we specify the `axis` as a `mg.Tensor` or `np.ndarray`? do we want to allow that?

src/mygrad/nnet/activations/softmax.py
rsokl: oh... ha. You are right

src/mygrad/nnet/activations/softmax.py
rsokl: At that point I might as well put `Sequence[int]`. I think I'll just keep it simple though

src/mygrad/nnet/activations/softmax.py
rsokl, line 17:
```suggestion
        x = np.ones_like(x)
```

PR #269, opened 2020-08-03
Improve focal loss

### LINKED ISSUE #256: Write uber-wrapper tests for focal loss

rsokl:
(accidentally branched off of softmax branch)

The derivative of focal-loss is finite for gamma < 1, p -> 1. This was not apparent when I derived softmax-focal loss, because the inclusion of the softmax lead to cancellations that made the application L'Hôpital's rule far more subtle.

Thus this PR provides backprop through "bare" focal loss, handling all of the limits / edge cases (gamma = 0, gamma = 1, gamma < 1 & p -> 1) in this simpler form. softmax-focal loss then simply calls into this to avoid making things unnecessarily complicated.

This PR also hardens the testing for focal loss by using the uber test wrappers. It also fixes a bug where specifying `constant` for the `focal_loss` function didn't properly set the output constant.

closes #256

### DIFF
--- a/src/mygrad/nnet/losses/focal_loss.py
+++ b/src/mygrad/nnet/losses/focal_loss.py
@@ -1,6 +1,7 @@
 import numpy as np
 
-from mygrad import Tensor, log
+from mygrad import Tensor
+from mygrad.nnet.activations import softmax
 from mygrad.operation_base import Operation
 
 from ._utils import check_loss_inputs
@@ -8,7 +9,7 @@ from ._utils import check_loss_inputs
 __all__ = ["softmax_focal_loss", "focal_loss"]
 
 
-class SoftmaxFocalLoss(Operation):
+class FocalLoss(Operation):
     r""" Returns the per-datum focal loss as described in https://arxiv.org/abs/1708.02002
     which is given by -ɑ(1-p)ˠlog(p).
 
@@ -29,11 +30,11 @@ class SoftmaxFocalLoss(Operation):
 
     scalar_only = True
 
-    def __call__(self, scores, targets, alpha, gamma):
+    def __call__(self, class_probs, targets, alpha, gamma):
         """
         Parameters
         ----------
-        scores : mygrad.Tensor, shape=(N, C)
+        class_probs : mygrad.Tensor, shape=(N, C)
             The C class scores for each of the N pieces of data.
 
         targets : Union[mygrad.Tensor, Sequence[int]], shape=(N,)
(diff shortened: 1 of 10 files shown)


src/mygrad/nnet/losses/focal_loss.py
```diff
@@ -50,54 +51,70 @@ def __call__(self, scores, targets, alpha, gamma):
...
     Parameters
     ----------
-    scores : mygrad.Tensor, shape=(N, C)
-        The C class scores for each of the N pieces of data.
+    class_probs : mygrad.Tensor, shape=(N, C)
+        The C class probabilities for each of the N pieces of data.
+        Each value is expected to lie on (0, 1]
 
-    targets : array_like, shape=(N,)
+    targets : Sequence[int], shape=(N,)
```
davidmascharka, line 109: can these be consistent?

src/mygrad/nnet/losses/focal_loss.py
```diff
@@ -124,27 +142,32 @@ def softmax_focal_loss(scores, targets, *, alpha=1, gamma=0, constant=False):
...
+    Applies the softmax normalization to the input scores before computing the
+    per-datum focal loss.
 
     Parameters
     ----------
     scores : mygrad.Tensor, shape=(N, C)
         The C class scores for each of the N pieces of data.
 
-    targets : Sequence[int], shape=(N,)
+    targets : array_like, shape=(N,)
```
davidmascharka, line 162: :thinking:

src/mygrad/nnet/losses/focal_loss.py
```diff
@@ -50,54 +51,70 @@ def __call__(self, scores, targets, alpha, gamma):
...
-        self.back *= alpha
-        return loss
+        one_m_pc_gamma = one_m_pc ** gamma
+        loss = -(alpha * one_m_pc_gamma * log_pc)
+
+        self.back = np.zeros(class_probs.shape, dtype=np.float64)
+
+        if np.isclose(gamma, 0, atol=1e-15):
+            self.back[self.label_locs] -= alpha / pc
+            return loss
```
davidmascharka, line 76: could we remove this and just de-dent the other `return` so that there's only one exit point to the function?

```suggestion
```

src/mygrad/nnet/losses/focal_loss.py
```diff
@@ -50,54 +51,70 @@ def __call__(self, scores, targets, alpha, gamma):
...
+                term2 = np.zeros(pc.shape, dtype=np.float64)
+                pc_not_1 = ~np.isclose(one_m_pc, 0, atol=1e-15)
+                term2[pc_not_1] = (
+                    -gamma * one_m_pc[pc_not_1] ** (gamma - 1) * log_pc[pc_not_1]
+                )
+            else:
+                term2 = -gamma * one_m_pc ** (gamma - 1) * log_pc
+
+            self.back[self.label_locs] -= alpha * (term1 + term2)
+            return loss
```
davidmascharka, line 93:
```suggestion
        return loss
```
to account for the above

tests/nnet/losses/test_focal_loss.py
```diff
@@ -1,13 +1,47 @@
...
+    return -alpha * np.clip(1 - pc, a_min=0, a_max=1) ** gamma * np.log(pc)
+
+
+def numpy_softmax_focal_loss(
+    scores: np.ndarray, targets: np.ndarray, alpha: float, gamma: float
+) -> np.ndarray:
+    if isinstance(targets, mg.Tensor):
+        targets = targets.data
+    scores = softmax(scores).data
+    rows = np.arange((len(scores)))
```
davidmascharka, line 29: unnecessary extra parens?
```suggestion
    rows = np.arange(len(scores))
```

tests/nnet/losses/test_focal_loss.py
```diff
@@ -25,76 +59,73 @@ def test_input_validation(fn, data, labels):
...
+@backprop_test_factory(
+    mygrad_func=focal_loss,
+    true_func=numpy_focal_loss,
+    index_to_arr_shapes={0: hnp.array_shapes(min_dims=2, max_dims=2)},
+    index_to_bnds={0: (1e-14, 1)},
+    num_arrays=1,
+    kwargs=dict(
+        targets=lambda scores: targets(scores=scores),
+        alpha=lambda scores: st.floats(-2, 2),
+        gamma=lambda scores: st.floats(0, 10) | st.sampled_from([0.0, 1.0]),
```
davidmascharka, line 93: I'm assuming the `| st.sampled_from([0, 1])` is just to increase the chance of those exact values being exercised? Is it possible to provide them as `@example`s?

src/mygrad/nnet/losses/focal_loss.py
rsokl: Agreed that this return needs to be made more conspicuous - I opted to simply remove the subsequent `else` and de-dent everything else. Basically, this is a special early return condition

tests/nnet/losses/test_focal_loss.py
rsokl: Yeah that just helps to ensure those edge cases get exercises.

I wish there was a way to do "partial" examples like this, but there isn't. I believe you have to provide examples for everything in the signature.

I'll have to check with the hypothesis devs to see if they are interested in changing this.

PR #270, opened 2020-08-10
Fully replace add.at in getitem

petarmhg:
This aims to fully phase out the use of `add.at` in the getitem backprop (as started in #265 ). The tests for `getitem`, however, cannot currently check all forms of advanced indexing, and I found a few simple cases in checking the function by hand that broke the code (those issues should be fixed).

~~I am going to work on making the tests for `getitem` more robust (e.g. check for slices, integer arrays, and boolean arrays in the same index). While these types of indexing will, for the most part, never be used, it's be nice to catch and correct any errors.~~ (see #271 for progress on this)

I also still need to do a better performance comparison of this implementation vs `add.at`. From my initial checks I noticed that for small arrays, `add.at` is slightly faster (a few μs) because of the additional logic here. For larger arrays, or when many values are being indexed (as is common when a slice is included), this tends to run a good bit faster. Again I will need to do look more in detail at how the runtime scales for this and for `add.at`

### DIFF
--- a/src/mygrad/tensor_core_ops/indexing.py
+++ b/src/mygrad/tensor_core_ops/indexing.py
@@ -73,7 +73,7 @@ class GetItem(Operation):
         a = self.variables[index]
         inds = self.index
 
-        if self._used_distinct_indices:
+        if self._used_distinct_indices or grad.size == 0:
             out = np.zeros_like(a.data)
             out[inds] += grad
         else:
@@ -88,7 +88,7 @@ class GetItem(Operation):
             ell_ind = -1
 
             # used to determine whether all integer array indices are next to one another or not
-            # and if not, where indices must be inserted for broadcast compatability
+            # and if not, where indices must be inserted for broadcast compatibility
             pos_int_arr = []
 
             # tally objects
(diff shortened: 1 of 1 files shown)


rsokl:
This would be a welcome improvement! Perhaps you could open a separate PR to harden the tests for `__getitem__` so that we can verify that they don't catch any bugs in the current version of mygrad, and then this PR can be responsible for improving the method itself.

rsokl:
@petarmhg given the myriad of edge-cases happening here, it would be useful to include some manual tests - so that we don't have to hope that your strategy hits all of the edge cases each time the test suite runs. This will probably lead to flaky coverage, given how exotic some of these indices are.

rsokl:
>I also still need to do a better performance comparison of this implementation vs add.at. From my initial checks I noticed that for small arrays, add.at is slightly faster (a few μs) because of the additional logic here. For larger arrays, or when many values are being indexed (as is common when a slice is included), this tends to run a good bit faster. Again I will need to do look more in detail at how the runtime scales for this and for add.at

This analysis will be pretty important to whether or not we end up going with this. The cognitive overhead here is pretty large - this went from a couple of lines of code to one of the most complicated pieces of code in MyGrad!

PR #271, opened 2020-08-11
Improve getitem tests

petarmhg:
Tests `getitem` on indices with arbitrary (valid) combinations of ints, slices, integer arrays, boolean arrays, etc

### DIFF
--- a/tests/custom_strategies/__init__.py
+++ b/tests/custom_strategies/__init__.py
@@ -14,6 +14,7 @@ from numpy import ndarray
 
 __all__ = [
     "adv_integer_index",
+    "arbitrary_indices",
     "basic_indices",
     "broadcastable_shapes",
     "choices",
@@ -381,6 +382,8 @@ def arbitrary_indices(draw, shape: Tuple[int]):
     `draw` is a parameter reserved by hypothesis, and should not be specified
     by the user.
 
+    When given a shape with a 0-dimensional axis, only a basic index will be returned.
+
     Returns
     -------
     hypothesis.searchstrategy.SearchStrategy[Tuple[Union[int, slice, Ellipsis, NoneType, numpy.ndarray], ...]]
@@ -412,7 +415,7 @@ def arbitrary_indices(draw, shape: Tuple[int]):
         return draw(hnp.basic_indices(shape=shape, allow_newaxis=True))
 
     shape_inds = list(range(len(shape)))
-    index = []  # stores tuples of (dim, indexing object)
+    index = []  # stores tuples of (axis, indexing object)
 
     # add integers, slices
     basic_inds = sorted(draw(st.lists(st.sampled_from(shape_inds), unique=True)))
@@ -429,7 +432,7 @@ def arbitrary_indices(draw, shape: Tuple[int]):
 
         index += [tup for tup in zip(basic_inds, basics)]
 
-        # will not necessarily index all dims from basic_inds as
+        # will not necessarily index all axes from basic_inds as
         # `basic_indices` can return indices with omitted trailing slices
         # so only remove dimensions directly indexed into
         for i in basic_inds[: len(basics)]:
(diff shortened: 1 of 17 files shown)


tests/tensor_ops/test_getitem.py
```diff
@@ -220,3 +241,27 @@ def test_getitem_basic_w_adv_fwdprop():
...
+def test_getitem_arbitraryindex_fwdprop():
+    pass
+
+
+@settings(deadline=None)
+@backprop_test_factory(
+    mygrad_func=get_item,
+    true_func=get_item,
+    num_arrays=1,
+    index_to_arr_shapes={0: hnp.array_shapes(max_side=4, max_dims=8)},
```
rsokl, line 262:
```suggestion
    index_to_arr_shapes={0: hnp.array_shapes(min_side=0, max_side=4, min_dims=0, max_dims=5)},
```

tests/tensor_ops/test_getitem.py
```diff
@@ -220,3 +241,27 @@ def test_getitem_basic_w_adv_fwdprop():
 )
 def test_getitem_basic_w_adv_bkprop():
     pass
+
+
+@fwdprop_test_factory(
+    mygrad_func=get_item,
+    true_func=get_item,
+    num_arrays=1,
+    index_to_arr_shapes={0: hnp.array_shapes(max_side=4, max_dims=8)},
```
rsokl, line 250: Note that running this locally revealed the mygrad bug: https://github.com/rsokl/MyGrad/issues/272

```suggestion
    index_to_arr_shapes={0: hnp.array_shapes(min_side=0, max_side=4, min_dims=0, max_dims=5)},
```

tests/tensor_ops/test_getitem.py
```diff
@@ -220,3 +241,27 @@ def test_getitem_basic_w_adv_fwdprop():
...
+    kwargs=dict(index=arb_index_wrap),
+)
+def test_getitem_arbitraryindex_fwdprop():
+    pass
+
+
+@settings(deadline=None)
```
rsokl, line 257: This is a powerful test - it is worth letting hypothesis do a lot of searching for edge cases
```suggestion
@settings(deadline=None, max_examples=500)
```

tests/custom_strategies/__init__.py
```diff
@@ -358,3 +360,158 @@ def valid_shapes(
...
+        Examples
+        --------
+        >>> group_continuous_integers([1, 3, 4, 5, 7, 8])
+        [(1,), (3, 4, 5), (7, 8)]
+        """
+        return [
+            tuple(map(itemgetter(1), g))
+            for k, g in groupby(enumerate(ls), lambda x: x[0] - x[1])
+        ]
+
```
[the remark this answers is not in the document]
petarmhg: Ahh nice catch. In the case of empty arrays, I think this fix will make it so that the case

```
np.array([])[[]]
```

could never be checked (though for what reason this would ever need to be used, I don't know). Since the `integer_array_indices` strategy doesn't handle dim-0 axes, I can add a small test to `test_getitem` to check for this

tests/tensor_ops/test_getitem.py
```diff
@@ -48,6 +47,19 @@ def adv_index_bool_wrap(*arrs):
...
+    return arbitrary_indices(arrs[0].shape)
+
+
+def test_index_empty():
+    a = Tensor([])
+    assert a[[]].shape == (0,)
+
+    a.sum().backward()
+    # since a is empty, a.grad should be empty and same shape
+    assert np.allclose(a, a.grad)
```
rsokl, line 60: `np.allclose` isn't strict about shapes with empty arrays

```python
>>> np.allclose(np.array([[]]) , np.array([]))
True
```
Let's check the shapes explicitly

```suggestion
def test_index_empty():
    a = mg.Tensor([])
    b = a[[]]
    
    assert b.shape == (0,)

    b.sum().backward()
    # since a is empty, a.grad should be empty and same shape
    assert a.shape == a.grad.shape
    assert b.shape == b.grad.shape
```

tests/tensor_ops/test_getitem.py
```diff
@@ -48,6 +47,22 @@ def adv_index_bool_wrap(*arrs):
     return hnp.arrays(shape=arrs[0].shape, dtype=bool)
 
 
+def arb_index_wrap(*arrs):
+    return arbitrary_indices(arrs[0].shape)
+
+
+def test_index_empty():
+    a = mg.Tensor([])
```
rsokl, line 55: Ugh sorry. That was my fault.

```suggestion
    a = Tensor([])
```

rsokl:
🥳

PR #273, opened 2020-08-14
permit indexing into 0d arrays

### LINKED ISSUE #272: 0D Tensors raise on get-item instead of on iteration
In an attempt to fix #210 vis #216, we accidentally blocked off indexing into 0D tensors

The following should be legal:

```python
>>> import mygrad as mg
>>> mg.Tensor(3)[None]
Tensor([3])
```

rsokl:
closes #272

### DIFF
--- a/tests/tensor_ops/test_iter.py
+++ b/tests/tensor_ops/test_iter.py
@@ -1,0 +1,45 @@
+import hypothesis.extra.numpy as hnp
+import pytest
+
+from mygrad.tensor_base import Tensor
+
+from ..wrappers.uber import backprop_test_factory, fwdprop_test_factory
+
+
+# Test https://github.com/rsokl/MyGrad/issues/210
+def test_0d_iter():
+    x = Tensor(3)
+    with pytest.raises(TypeError):
+        sum(x)
+
+
+def _sum(x, constant=False):
+    out = sum(x)
+    if isinstance(x, Tensor):
+        if not isinstance(out, Tensor):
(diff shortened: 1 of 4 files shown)


tests/tensor_ops/test_iter.py
```diff
@@ -0,0 +1,45 @@
...
+)
+def test_builtin_sum_fwd():
+    pass
+
+
+@backprop_test_factory(
+    mygrad_func=_sum,
+    true_func=_sum,
+    num_arrays=1,
+    index_to_arr_shapes={0: hnp.array_shapes(min_dims=1, min_side=1)},
```
petarmhg, line 41: Is the `min_side=1` is because `sum(mg.Tensor([]))` yields `0`?

Wouldn't it also exclude from the test valid cases like
```
>>> a = mg.arange(10*10*10).reshape(10,10,10)
>>> b = sum(a[:, slice(0, 1, -1), :]) # shape (10, 0, 10)
>>> b.backward()
```

(anytime when the 0-dim is anywhere except on the 0th axis)

rsokl: Yeah, this is a tricky case where `sum` returns `0` (an int, not a tensor), so we can't even invoke backprop.. I don't think there is any way to avoid this, since the tensor is actually an empty iterable and it is `sum` that returns 0.

rsokl: Oh dang... really nice catch. You are right about that!

PR #276, opened 2020-08-14
Demonstrate that python3 wants to warn you not to override __eq__ withou __hash__

aslvrstn:
Normally python would have made `Tensor` unhashable, but because this is being monkeypatched, it wasn't able to.

You would think this PR is a noop, but it'll cause a bunch of 'unhashable type: Tensor' errors

Ryan Edit:

Any class with `__eq__` defined will have that method be used to check hash-table collisions. `Tensor.__eq__` is based on its underlying numpy array, which is mutable. Also, this method returns an array, not a boolean – making it even more inappropriate. I.e. `Tensor` should have been unhashable.

`Tensor` _was_ hashable – and we relied on this hashability for doing BFS traversals of the computational graph – but this was a fluke. It was because `__eq__` and `__neq__` was getting monkeypatched after the `Tensor` definition. So basically, Python was falling back on its default behavior for providing bare classes with hashes: relying on object-id.

This PR replaces any reliance on hashing Tensor with hashing that tensor's id.   

☠️ relevant explanation  https://hynek.me/articles/hashes-and-equality/

and thanks to Alex for finding / explaining  all of this. I am rehashing it here for posterity.

### DIFF
--- a/src/mygrad/operation_base.py
+++ b/src/mygrad/operation_base.py
@@ -151,8 +151,10 @@ class Operation:
         # these by the node itself, since Tensors are unhashable, but by its `id`.
         visited = set()
         for var in (
-            i for i in self.variables if not i.constant and i.creator is not None and id(i) not in visited
+            i for i in self.variables if not i.constant and i.creator is not None
         ):
+            if id(var) in visited:
+                continue
             visited.add(id(var))
             var._accum_ops.add(self)
             var._backward(graph=graph)
--- a/src/mygrad/operation_base.py
+++ b/src/mygrad/operation_base.py
@@ -147,9 +147,13 @@ class Operation:
                         var.grad += backed_grad
                     else:
                         var.grad += _reduction(backed_grad, var.shape)
+        # Avoid visiting the same node multiple times. Note that we don't store
+        # these by the node itself, since Tensors are unhashable, but by its `id`.
+        visited = set()
         for var in (
-            i for i in self.variables if not i.constant and i.creator is not None
+            i for i in self.variables if not i.constant and i.creator is not None and id(i) not in visited
         ):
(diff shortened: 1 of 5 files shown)


rsokl:
☠️

src/mygrad/tensor_base.py
```diff
@@ -881,6 +882,12 @@ def T(self):
         """
         return self._op(Tensor_Transpose_Property, self)
 
+    def __eq__(self):
```
rsokl, line 885: This is the wrong signature

src/mygrad/operation_base.py
```diff
@@ -152,9 +152,13 @@ def backward(self, grad, *, graph, _reduction=None, **kwargs):
                         var.grad += backed_grad
                     else:
                         var.grad += _reduction(backed_grad, var.shape)
+        # Avoid visiting the same node multiple times. Note that we don't store
+        # these by the node itself, since Tensors are unhashable, but by its `id`.
+        visited = set()
         for var in (
-            i for i in self.variables if not i.constant and i.creator is not None
+            i for i in self.variables if not i.constant and i.creator is not None and id(i) not in visited
```
rsokl, line 159: I think this breaks `x * x` <- the same tensor is used twice. If the test suite doesn't catch this, then I definitely need to add tests for it.

aslvrstn: They don't! Everything's passing locally. Why would this break it? It seems equivalent to the world before.

rsokl:
Do you think we should add a regression test where we try to hash a tensor?

aslvrstn:
Is it not possible to respond to a comment that's not a comment on a line?

Anyway, added a test.

PR #278, opened 2020-08-15
remove ability to set Tensor.constant

rsokl:
Reverts #267 

There were tons of problems with setting Tensor.constant after it was already involved in a computational graph. The state-based testing found this, but needed to traverse deeper to find the problems consistently.

E.g.

```python
v1 = Tensor(0.)
v2 = +v1
v3 = Tensor(0.)
v4 = v3 - v2
v4.constant = True
v5 = v4 + v2

v5.backward()
```

would fail to clear `v2._accum_ops` during backprop. Another fail case: https://travis-ci.com/github/rsokl/MyGrad/jobs/372350422#L395

I considered only partially restricting the user's ability to set constant, but it just created too many headaches.

As an alternative, users can now use `Tensor.copy(constant)` to a similar, but sanitized, effect

### DIFF
--- a/src/mygrad/tensor_base.py
+++ b/src/mygrad/tensor_base.py
@@ -712,7 +712,7 @@ class Tensor:
         """
         return self.copy()
 
-    def copy(self, constant: Optional[None] = None) -> "Tensor":
+    def copy(self, constant: Optional[bool] = None) -> "Tensor":
         """ Produces a copy of ``self`` with ``copy.creator=None``.
 
         Copies of the underlying numpy data array and gradient array are created.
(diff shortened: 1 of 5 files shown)


src/mygrad/tensor_base.py
```diff
@@ -716,20 +710,20 @@ def __copy__(self):
...
-            _creator=None,
-            constant=self.constant,
-            _scalar_only=self._scalar_only,
-        )
-        copy.grad = np.copy(self.grad) if self.grad is not None else None
-        return copy
+        return self.copy()
 
-    def copy(self):
+    def copy(self, constant: Optional[None] = None) -> "Tensor":
```
rsokl, line 715:
```suggestion
    def copy(self, constant: Optional[bool] = None) -> "Tensor":
```

PR #279, opened 2020-08-15
Add asarray

rsokl:
The first of many convenience functions for interchanging between arrays and tensors

Also adds `tensors` hypothesis strategy

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [flake8]
-exclude = docs/*, src/mygrad/__init__.py
+exclude = docs/*, src/mygrad/__init__.py, src/mygrad/numpy_compat/__init__.py
 
 
 [versioneer]
--- a/src/mygrad/numpy_compat/array_creation.py
+++ b/src/mygrad/numpy_compat/array_creation.py
@@ -13,7 +13,7 @@ def asarray(a, dtype=None, order=None) -> np.ndarray:
     Parameters
     ----------
     a : array_like
-        Input data, in any form that can be converted to an array.  This
+        Input data, in any form - including a mygrad tensor - that can be converted to an array. This
         includes lists, lists of tuples, tuples, tuples of tuples, tuples
         of lists and ndarrays.
 
--- a/src/mygrad/numpy_compat/array_creation.py
+++ b/src/mygrad/numpy_compat/array_creation.py
@@ -36,15 +36,16 @@ def asarray(a, dtype=None, order=None) -> np.ndarray:
     --------
     Convert a list into an array:
 
+    >>> import mygrad as mg
     >>> a = [1, 2]
-    >>> np.asarray(a)
+    >>> mg.asarray(a)
     array([1, 2])
 
     Convert a tensor into an array. No copy of the
     underlying numpy array is created:
 
     >>> t = mg.Tensor([1, 2.])
-    >>> np.asarray(t)
+    >>> mg.asarray(t)
     array([1., 2.])
     >>> t.data is np.asarray(t))
     True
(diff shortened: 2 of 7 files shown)


src/mygrad/numpy_compat/array_creation.py
```diff
@@ -0,0 +1,79 @@
...
+    """Convert the input to an array.
+
+    This docstring is adapted from that of ``numpy.asarray``
+
+    Parameters
+    ----------
+    a : array_like
+        Input data, in any form that can be converted to an array.  This
+        includes lists, lists of tuples, tuples, tuples of tuples, tuples
+        of lists and ndarrays.
```
rsokl, line 18:
```suggestion
        of lists, ndarrays, and mygrad-tensors.
```

src/mygrad/numpy_compat/array_creation.py
```diff
@@ -0,0 +1,79 @@
...
+
+def asarray(a, dtype=None, order=None) -> np.ndarray:
+    """Convert the input to an array.
+
+    This docstring is adapted from that of ``numpy.asarray``
+
+    Parameters
+    ----------
+    a : array_like
+        Input data, in any form that can be converted to an array.  This
```
rsokl, line 16:
```suggestion
        Input data, in any form - including a mygrad tensor - that can be converted to an array. This
```

tests/custom_strategies/test_strategies.py
```diff
@@ -196,3 +199,75 @@ def test_arbitrary_indices_strategy(a, data):
...
     # rules, numpy will raise an error
     a[index]
+
+
+@pytest.mark.parametrize("constant", [st.booleans(), None])
+def test_tensors_handles_constant_strat(constant):
+    constants = []
+    kwargs = dict(dtype=np.int8, shape=(2, 3))
+    if constant is not None:
+        kwargs["constant"] = constant
```
petarmhg, line 209: What is the purpose of the `None`? If `constant` is not specified in the strategy, it uses `st.booleans` anyway. Wouldn't it make more sense to use `[True, False, None]` in the `parametrize` decorator?

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [flake8]
-exclude = docs/*, src/mygrad/__init__.py
+exclude = docs/*, src/mygrad/__init__.py, src/mygrad/numpy_compat/__init__.py
```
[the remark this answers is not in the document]
rsokl: It was mad that I imported `asarray` into that namespace without using it there.

tests/custom_strategies/test_strategies.py
rsokl: This just ensures that `None` has the same behavior as specifying `st.booleans()`. The static constant  cases are tested below.

rsokl:
> Could we put up an issue to keep track of which functions have been added and which have not?

Yeah, this would be useful. I don't have a definitive list in mind. `astensor` will definitely be one, but I am still coming up with ideas for others. I should have made `asarray` and `astensor` long ago.. is would have saved us so many if-clauses :/

tests/custom_strategies/__init__.py
```diff
@@ -70,6 +73,155 @@ def _check_min_max(min_val, min_dim, max_dim, param_name, max_val=None):
...
+
+    constant : Union[bool, st.SearchStrategy[bool]]
+        Specifies ``tensor.constant``. Default is :func:`~hypothesis.strategies.booleans`
+
+    include_grad: bool, optional (default=False)
+        If True, draw an array to be set to `tensor.grad`
+
+    grad_dtype: Optional[Any]
+        The datatype of the drawn gradient. Defaults to the dtype of the host tensor.
+
```
rsokl, line 134:
```suggestion
    grad_dtype: Optional[Any]
        The datatype of the drawn gradient. Defaults to the dtype of the host tensor.
        Specifying ``grad_dtype``, while ``include_grad`` is False, will raise an error.

```

tests/custom_strategies/__init__.py
```diff
@@ -70,6 +73,156 @@ def _check_min_max(min_val, min_dim, max_dim, param_name, max_val=None):
...
+        If True, draw an array to be set to `tensor.grad`
+
+    grad_dtype: Optional[Any]
+        The datatype of the drawn gradient. Defaults to the dtype of the host tensor.
+        Specifying ``grad_dtype``, while ``include_grad`` is False, will raise an error.
+
+    grad_elements_bounds: Optional[Tuple[int, int]]
+        The min and max bounds used to draw the gradient's elements.
+        Defaults to (-10, 10)
+
```
rsokl, line 139:
```suggestion
    grad_elements_bounds: Optional[Tuple[int, int]]
        The min and max bounds used to draw the gradient's elements.
        Defaults to (-10, 10)
        Specifying ``grad_element_bounds``, while ``include_grad`` is False, will raise an error.

```

tests/custom_strategies/test_strategies.py
petarmhg: Ahh I see :+1:

rsokl:
@petarmhg added the docs - should be good to go if that is everything

PR #281, opened 2020-08-17
Improve setitem tests

rsokl:
Consolidates a bunch of boilerplate code into a wrapper.

Tests against `arbitrary_indices` strategy

### DIFF
--- a/tests/tensor_ops/test_setitem.py
+++ b/tests/tensor_ops/test_setitem.py
@@ -92,15 +92,15 @@ class set_item_test_factory:
             note("x[index]: {}".format(o))
             y = data.draw(self.value_strat(o), label="y",)
 
-            x0 = np.copy(x)
-            y0 = np.copy(y)
+            numpy_x = np.copy(x)
+            numpy_y = np.copy(y)
 
-            x_arr = Tensor(np.copy(x))
-            y_arr = Tensor(np.copy(y))
-            x1_arr = +x_arr
+            mygrad_x = Tensor(np.copy(x))
+            mygrad_y = Tensor(np.copy(y))
+            mygrad_x1 = +mygrad_x
 
             try:
-                x0[index] = y0  # don't permit invalid set-items
+                numpy_x[index] = numpy_y  # don't permit invalid set-items
             except Exception:
                 assume(False)
                 return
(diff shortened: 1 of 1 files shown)


tests/tensor_ops/test_setitem.py
```diff
@@ -11,6 +14,7 @@
     _is_bool_array_index,
     _is_int_array_index,
 )
+from tests.custom_strategies import arbitrary_indices
 
 from ..custom_strategies import adv_integer_index, basic_indices, broadcastable_shapes
```
petarmhg, line 19: any reason you separate the imports from `custom_strategies`?

tests/tensor_ops/test_setitem.py
```diff
@@ -59,6 +63,71 @@ def setitem(x, y, index):
...
+
+            index = data.draw(self.index_strat(x), label="index")
+
+            try:
+                o = np.asarray(x[index])
+            except IndexError:
+                assume(False)
+                return
+
+            note("x[index]: {}".format(o))
```
petarmhg, line 92:
```suggestion
            note(f"x[index]: {o}")
```

tests/tensor_ops/test_setitem.py
```diff
@@ -59,6 +63,71 @@ def setitem(x, y, index):
...
+        self,
+        array_strat: st.SearchStrategy[np.ndarray],
+        index_strat: Callable[[np.ndarray], st.SearchStrategy],
+        value_strat: Callable[[np.ndarray], st.SearchStrategy[np.ndarray]],
+    ):
+        self.array_strat = array_strat
+        self.index_strat = index_strat
+        self.value_strat = value_strat
+
+    def __call__(self, f):
```
petarmhg, line 77: I assume it would be too messy to type-hint `f` here

tests/tensor_ops/test_setitem.py
rsokl: Nope! Fixed

tests/tensor_ops/test_setitem.py
rsokl: Good call. The hint here isn't so bad since these functions don't actually accept or return anything:

```python
def __call__(self, f: Callable) -> Callable[[], None]:
```

rsokl:
I am closing this in favor of #292  because scope-creep turned this PR into something much more than it was supposed to be. I will open a separate PR to catch the mutation issues.

PR #289, opened 2020-08-23
add tests for augmented updates

rsokl:
I was trying to parameterize this and invoke e.g. `getattr(x, "__imul__")`, but this seemed to not work for some reason.. also it was so gross to read/write

### DIFF
--- a/tests/tensor_base/test_augmented_updates.py
+++ b/tests/tensor_base/test_augmented_updates.py
@@ -3,6 +3,9 @@ import mygrad as mg
 import numpy as np
 
 
+# TODO: These tests need to be modified to check that tensor-identity
+# is preserved by the augmented update once #280 is resolved
+# https://github.com/rsokl/MyGrad/pull/280/
 def test_augmented_multiply():
     a = np.arange(9.0).reshape(3, 3)
     expected_grad = np.where(a < 4, 2, 1)
--- a/tests/tensor_base/test_augmented_updates.py
+++ b/tests/tensor_base/test_augmented_updates.py
@@ -13,6 +13,9 @@ def test_augmented_multiply():
 
     t1 = mg.arange(9.0).reshape(3, 3)
     y = mg.full(t1[t1 < 4].shape, 2.0)
+    # creates a new node in the graph so that `t1` can
+    # effectively mirror `t2` before it was mutated, and
+    # thus store the gradient downstream of the augmented update
     t2 = +t1
     t2[t2 < 4] *= y
     t2.sum().backward()
--- a/tests/tensor_base/test_augmented_updates.py
+++ b/tests/tensor_base/test_augmented_updates.py
@@ -117,10 +117,10 @@ def test_augmented_power():
 
 def test_augmented_matmul():
     a = np.arange(9.0).reshape(3, 3)
-    t = np.arange(9.0).reshape(3, 3)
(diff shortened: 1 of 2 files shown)


tests/tensor_base/test_augmented_updates.py
```diff
@@ -0,0 +1,111 @@
...
+import numpy as np
+
+
+# TODO: These tests need to be modified to check that tensor-identity
+# is preserved by the augmented update once #280 is resolved
+# https://github.com/rsokl/MyGrad/pull/280/
+def test_augmented_multiply():
+    a = np.arange(9.0).reshape(3, 3)
+    expected_grad = np.where(a < 4, 2, 1)
+    a[a < 4] *= 2
```
aslvrstn, line 12: Can you swap lines 11 and 12 so it's clear why the gradient should be expected after you do the operation rather than before?

tests/tensor_base/test_augmented_updates.py
```diff
@@ -0,0 +1,111 @@
...
+
+    t2.null_gradients()
+    assert t2.grad is None and len(t2._ops) == 1 and not t2._accum_ops
+    assert t1.grad is None and not t1._ops and not t1._accum_ops
+    assert y.grad is None and not y._ops and not y._accum_ops
+
+
+def test_augmented_add():
+    a = np.arange(9.0).reshape(3, 3)
+    expected_grad = np.where(a < 4, 1, 1)
```
aslvrstn, line 32: and here

tests/tensor_base/test_augmented_updates.py
```diff
@@ -0,0 +1,111 @@
+from numpy.testing import assert_allclose
+import mygrad as mg
+import numpy as np
+
+
+# TODO: These tests need to be modified to check that tensor-identity
+# is preserved by the augmented update once #280 is resolved
+# https://github.com/rsokl/MyGrad/pull/280/
+def test_augmented_multiply():
```
aslvrstn, line 9: Missing imatmul?

tests/tensor_base/test_augmented_updates.py
```diff
@@ -0,0 +1,111 @@
...
+    a = np.arange(9.0).reshape(3, 3)
+    expected_grad = np.where(a < 4, 2, 1)
+    a[a < 4] *= 2
+
+    t1 = mg.arange(9.0).reshape(3, 3)
+    y = mg.full(t1[t1 < 4].shape, 2.0)
+    t2 = +t1
```
aslvrstn, line 16: Probably worth adding a comment about how you want this to form a new node. I don't think that's idiomatic for anyone but you.

tests/tensor_base/test_augmented_updates.py
rsokl: `a` get's mutated in line 12, which is why I need to order it the way I did.

aslvrstn: Ha, right.

tests/tensor_base/test_augmented_updates.py
rsokl: Thanks for reminding me about this. I decided to block users from using `__imatmul__` until numpy supports it

tests/tensor_base/test_augmented_updates.py
rsokl: Good call. Thanks!

PR #290, opened 2020-08-23
Ensure backprop through pow special cases is covered by tests

rsokl:


aslvrstn:
Don't you special-case 1 and 2 as well? In https://github.com/rsokl/MyGrad/blob/master/src/mygrad/tensor_base.py#L758 ?

rsokl:
Yep, but since it is only special-cased for `Tensor.__pow__`, the tests are here: https://github.com/rsokl/MyGrad/blob/master/tests/tensor_base/test_pow_special_cases.py

PR #307, opened 2020-09-12
Add no graph mode

rsokl:
@aslvrstn @Zac-HD if you wouldn't mind taking a look at [`src/mygrad/_graph_tracking.py`](https://github.com/rsokl/MyGrad/blob/1253c23071d966165484678a1cff29b08ae4b35e/src/mygrad/_graph_tracking.py), I'd like to get your thoughts on if this implementation for a global flag is sensible, or if there are some details/precautions that I should heed

Adds `mygrad.no_autodiff`, which is a context manager and decorator that temporarily disables all of mygrad's graph-tracking and memory-locking logic.

```python
from mygrad import no_autodiff, arange

x = arange(10.)

with no_autodiff:
    y = 2 * x  # creates a constant tensor with no creator

# x._ops is empty - no operations were recorded
# x's memory is still writeable
```

```python
@no_autodiff
def forward_pass():
    # all mygrad operations in the function body will be performed
    # without any graph-tracking or memory locking
    pass
````

You can also effectively convert mygrad functions to their numpy counterparts now:

```python
import mygrad as mg
numpy_max = mg.no_autodiff(mg.max, to_numpy=True)
```

### DIFF
--- a/src/mygrad/_graph_tracking.py
+++ b/src/mygrad/_graph_tracking.py
@@ -1,9 +1,10 @@
 """
 Provides user interface for suspending computational graph tracking and back-propagation
 """
-
 from functools import wraps
 from typing import Callable
+
+import numpy as np
 
 __all__ = ["no_autodiff"]
 
@@ -26,14 +27,24 @@ class NoAutoDiff:
         global TRACK_GRAPH
         TRACK_GRAPH = True
 
-    def __call__(self, func: Callable) -> Callable:
-        """A decorated function will have graph-tracking suspended
-        during its execution."""
+    def __call__(self, func: Callable, to_numpy: bool = False) -> Callable:
+        """Decorates a function so that it will have graph-tracking suspended
+        during its execution.
+
+        Parameters
+        ----------
+        func : Callable
+            The function to be decorated
+
+        to_numpy : bool, optional (default=False)
+            If true, the output is assumed to be array-like and
+            will be cast to a numpy array"""
 
         @wraps(func)
         def wrapper(*args, **kwargs):
             with self:
-                return func(*args, **kwargs)
+                out = func(*args, **kwargs)
+            return out if not to_numpy else np.asarray(out)
 
         return wrapper
 
--- a/src/mygrad/_graph_tracking.py
+++ b/src/mygrad/_graph_tracking.py
@@ -13,19 +13,24 @@ __all__ = ["no_autodiff"]
 TRACK_GRAPH = True  # type: bool
 
 
-class NoAutoDiff:
+class _NoAutoDiff:
     """ Serves as a context manager and decorator for suspending
     all computational graph tracking."""
+
+    depth = 0
 
     def __enter__(self):
         """Suspends graph-tracking"""
         global TRACK_GRAPH
+        self.depth += 1
         TRACK_GRAPH = False
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         """Restores graph-tracking"""
         global TRACK_GRAPH
-        TRACK_GRAPH = True
+        self.depth -= 1
+
+        TRACK_GRAPH = self.depth == 0
 
     def __call__(self, func: Callable, to_numpy: bool = False) -> Callable:
         """Decorates a function so that it will have graph-tracking suspended
@@ -49,4 +54,4 @@ class NoAutoDiff:
         return wrapper
 
 
-no_autodiff = NoAutoDiff()
+no_autodiff = _NoAutoDiff()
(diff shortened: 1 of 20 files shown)


src/mygrad/operation_base.py
```diff
@@ -119,8 +119,10 @@ def backward(self, grad, *, graph, _reduction=None, **kwargs):
             if not var.constant:
                 if not var._ops:
                     raise InvalidBackprop(
-                        "Part of the computational graph containing "
-                        "this tensor was 'cleared' prior to backprop."
+                        f"Part of the computational graph containing "
+                        f"this tensor, {var}, was 'cleared' prior to backprop.\n"
+                        f"It is recommended that you clear all computational graphs "
+                        f"and restart your computation."
```
Zac-HD, line 125:
```suggestion
                        "Part of the computational graph containing "
                        f"this tensor, {var}, was 'cleared' prior to backprop.\n"
                        "It is recommended that you clear all computational graphs "
                        "and restart your computation."
```

src/mygrad/tensor_base.py
```diff
@@ -527,18 +528,55 @@ def _op(
...
+            for can_share_mem, var in zip(vars_can_share_mem, tensor_vars):
+                if can_share_mem and _is_view_of(parent=var, child=op_out):
+                    base = var if var.base is None else var.base
+                    assert not f.cannot_return_view, (
+                        f"{f} is marked as being unable to return a view, "
+                        f"however its output is a view of one of its inputs"
+                    )
+                    break
+            else:
+                base = None
```
Zac-HD, line 565: This loop makes me suspicious.

- You could define `vars_can_share_mem` immediately before use; for clarity and to avoid pointless work `if f.cannot_return_view`
- The assertion is redundant, because we're in the else-condition of a branch on that same test

src/mygrad/tensor_base.py
```diff
@@ -626,6 +648,9 @@ def backward(self, grad=None):
         >>> x.grad # df/dx = df/dw * dw/dx
         array(6.)
         """
+        if not _track.TRACK_GRAPH:
+            return
```
Zac-HD, line 652: Is it even valid to call `.backward()` in `no_autodiff` mode?

src/mygrad/_graph_tracking.py
```diff
@@ -0,0 +1,52 @@
...
+import numpy as np
+
+__all__ = ["no_autodiff"]
+
+
+# If `False`, suspends all computational graph tracking and backprop
+TRACK_GRAPH = True  # type: bool
+
+
+class NoAutoDiff:
```
Zac-HD, line 16:
```suggestion
class _NoAutoDiff:
```

Because I presume that this *really* shouldn't be touched from outside the module.

src/mygrad/_graph_tracking.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __enter__(self):
+        """Suspends graph-tracking"""
+        global TRACK_GRAPH
+        TRACK_GRAPH = False
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """Restores graph-tracking"""
+        global TRACK_GRAPH
+        TRACK_GRAPH = True
```
Zac-HD, line 28: You need to track the stack depth, or the following patterns are broken:

```python
with no_autodiff:
	# autodiff disabled here

    with no_autodiff:
		pass # disabled here too

	# whoops, it's re-enabled here!
```

and once you're dealing in decorated functions it's pretty easy to accidentally nest calls like this.

```suggestion
    def __enter__(self):
        """Suspends graph-tracking"""
        global TRACK_GRAPH
        TRACK_GRAPH = False
        depth += 1  # global variable, self.depth attribute, whatever

    def __exit__(self, exc_type, exc_val, exc_tb):
        """Restores graph-tracking"""
        global TRACK_GRAPH
        assert depth >= 1
        depth -= 1
        TRACK_GRAPH = depth == 0
```

src/mygrad/tensor_base.py
rsokl: My thoughts are that I want users to be able to basically just throw their training code under the `no_autodiff` context and have it be valid for performing inference. From this perspective, I think making `.backward()` simply be a no-op.

src/mygrad/_graph_tracking.py
rsokl: Great catch... I can only hope that I would have thought about this before this saw the light of day 😬

src/mygrad/tensor_base.py
rsokl: Good call on both fronts

src/mygrad/operation_base.py
rsokl: I like making these all f-strings, it is also how PyCharm auto-formats this

PR #320, opened 2020-12-29
Add iops

rsokl:
Adds support for augmented updates via arithmetic operators (e.g. `+=`)

As shown in #280, MyGrad use to simply "unroll" augmented updates:
```python
tensor += 2

# was the same as
tensor = tensor + 2
```

Meaning that the update did not actually occur in-place. These augmented updates now mirror NumPy's behavior.

Closes #280

### DIFF
--- a/src/mygrad/math/arithmetic/ops.py
+++ b/src/mygrad/math/arithmetic/ops.py
@@ -6,12 +6,19 @@ from mygrad.operation_base import BroadcastableOp, Operation
 
 __all__ = [
     "Add",
+    "IAdd",
     "Subtract",
+    "ISubstract",
     "Multiply",
+    "IMultiply",
     "Divide",
+    "IDivide",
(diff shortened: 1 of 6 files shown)


src/mygrad/math/arithmetic/ops.py
```diff
@@ -40,6 +47,35 @@ def backward_var(self, grad, index, **kwargs):
...
+
+        Returns
+        -------
+        inplace_target_data : numpy.ndarray
+
+        Notes
+        -----
+        Note that inplace_target will be replaced with its un-mutated
+        value by ``Tensor._inplace_op`` prior to backprop. Thus we need
+        not worry about caching inplace_targer and rewriting the backprop
```
aslvrstn, line 67: My only contribution is that you spelled this `inplace_targer` all over the place

PR #346, opened 2021-01-27
Tensor.__array__ should not accept a dtype parameter

rsokl:
It wasn't breaking anything but did not conform to the standard interface.


rsokl:
Docs got me! https://github.com/numpy/numpy/pull/18238

PR #379, opened 2021-07-23
Update default behavior of derivative of linalg.norm and abs at x=0

rsokl:
(I also smuggled in a fix for `mygrad.arange` that normalizes its interface with that of `numpy.arange`

Now both `mygrad.linalg.norm` and `mygrad.absolute`, by default, will return `0` wherever their derivatives would typically not defined (and would previously have returned `nan`).
Both of these functions can be passed `nan_to_num=False` to enable the previous, more-rigorous, behavior.

Before:

```python
>>> import mygrad as mg
>>> x = mg.tensor([-2.0, 0.0, 2.0])

>>> mg.absolute(x).backward()
>>> x.grad
array([-1., nan,  1.])
```

After:

```python
>>> import mygrad as mg
>>> x = mg.tensor([-2.0, 0.0, 2.0])

>>> mg.absolute(x).backward()
>>> x.grad
array([-1., 0.,  1.])

>>> mg.absolute(x, nan_to_num=False).backward()
>>> x.grad
array([-1., nan,  1.])
```

### DIFF
--- a/src/mygrad/linalg/funcs.py
+++ b/src/mygrad/linalg/funcs.py
@@ -118,8 +118,9 @@ def norm(
     >>> l2_norms
     Tensor([3.74165739, 1.        ])
 
-    When `nan_to_num=False` is specified the presence of the elementwise absolute values in the
-    norm operation means that zero-valued entries in any of input vectors will have an undefined derivative.
+    The presence of the elementwise absolute values in the norm operation means that zero-valued entries in any of 
+    input vectors have an undefined derivative. When `nan_to_num=False` is specified these derivatives will be reported
+    as `nan`, otherwise they will be made to be 0.0.
 
     >>> l2_norms = mg.linalg.norm(x, axis=1, ord=2, nan_to_num=True)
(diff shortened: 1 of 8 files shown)


src/mygrad/linalg/funcs.py
```diff
@@ -113,14 +118,25 @@ def norm(
     >>> l2_norms
     Tensor([3.74165739, 1.        ])
 
-    The presence of the elementwise absolute values in the norm means that zero-valued
-    entries in a vectors have an undefined derivative.
+    When `nan_to_num=False` is specified the presence of the elementwise absolute values in the
+    norm operation means that zero-valued entries in any of input vectors will have an undefined derivative.
```
rsokl, line 122:
```suggestion
    The presence of the elementwise absolute values in the norm operation means that zero-valued entries in any of 
    input vectors have an undefined derivative. When `nan_to_num=False` is specified these derivatives will be reported
    as `nan`, otherwise they will be made to be 0.0.
```

PR #381, opened 2021-08-04
Save and load func

kw-0:
Finished the save and load functions, thanks for the help by the way!


rsokl (commented):
Both `save` and `load` need to handle the cases where the tensor does not have a gradient associated with it.

rsokl:
I recommend importing your functions in a jupyter notebook and creating two test-cases: save and load a tensor with and without a gradient and see if the loaded tensor has the same data and gradient (if present) as the original tensor.

This sort of test is known as a "roundtrip" test and it is a very powerful one!

rsokl:
Oh actually it may be the case that your code does work when `grad` is `None`. Is the .npz file happy to store `None` in `"grad"`?

kw-0:
But I haven't actually tested that it saves None so I will try.

kw-0:
this doesn't actually fix the issue, sorry im a little distracted

kw-0:
Just got back and it works now!

src/mygrad/io.py
```diff
@@ -0,0 +1,53 @@
...
+    
+    Returns
+    -------
+    A tensor with the desired gradient data.
+    """
+    _tensor = np.load(tensor_filename, allow_pickle=True)
+
+    loaded_tensor = tensor(_tensor['data'])
+
+    if not _tensor['grad'].all():
```
rsokl, line 50: This doesn't do quite what you want. ` _tensor['grad']` checks to see if the array consists of all non-zero values. So this actually returns `False` if the gradient happens to contain any zeros, which means that any such gradient will not be loaded by this current code.

Given that we will want `save` to only include a `grad` field in the `.npz` file, we can do the following:

```python
with np.load(file") as opened_file:
        loaded_tensor = tensor(opened_file["data"])
    
    if "grad" in opened_file:
        loaded_tensor.backward(f["grad"])
```

The `with np.load(file) as opened_file` line opens the `.npz` file using a [context manager](https://www.pythonlikeyoumeanit.com/Module5_OddsAndEnds/WorkingWithFiles.html#Opening-Files). Within the context, `file` acts kind of like a dictionary whose keys are the array-names you use to save the `.npz` file, and whose values are the associated arrays. Thus `"grad" in file` will return `True` only if you included a `grad` field in the saved file.

Note that by avoiding saving `grad` when `grad` is `None`, we can avoid needing to specify `allow_pickle=True` during loading. It turns out that `allow_pickle=True` can create security vulnerabilities in software, so it is nice that we can avoid this.

src/mygrad/io.py
```diff
@@ -0,0 +1,53 @@
+import numpy as np
+import typing
+import mygrad as mg
+from mygrad.tensor_base import tensor, Tensor
+
+
+def save(filename : str, tensor : Tensor) -> None:
```
rsokl, line 7: Let's change `filename` to `file` so that we have "parity" with NumPy's `save` function

src/mygrad/io.py
```diff
@@ -0,0 +1,53 @@
...
+    A tensor with the desired gradient data.
+    """
+    _tensor = np.load(tensor_filename, allow_pickle=True)
+
+    loaded_tensor = tensor(_tensor['data'])
+
+    if not _tensor['grad'].all():
+        loaded_tensor.backward(_tensor['grad'])
+
+    return loaded_tensor
```
rsokl, line 53: For some reason it is customary to finish a file with a newline. This is why this red mark is here. It won't break the code but it is considered "proper formatting" (not sure why, really).

rsokl (changes requested):
Thanks for this PR @kw-0 ! There are a few changes that need to be make to get the behavior we are looking for. Let me know if you have any questions.

src/mygrad/io.py
kw-0: I like the way a newline at the end looks, also I forgot it in this commit. Oops.

rsokl (changes requested):
Nice! This is looking good.

I made some suggested changes, which you can accept in GitHub via your browser - GitHub will automatically commit those changes to your uploaded branch.

Another thing you will want to do is make these functions available at MyGrad's top-level namespace. I.e. we want people to be able to use `mygrad.save` and `mygrad.load`. So you will want to go to the file `mygrad/__init__.py` and add:

```python
from .io import load, save
```

This will make your functions available at the top-level namespace.

kw-0:
This is so cool! I've never done a PR on someone else's project, and I haven't worked on a real module before either. Thanks for the guidance thus far!

rsokl:
>This is so cool! I've never done a PR on someone else's project, and I haven't worked on a real module before either. Thanks for the guidance thus far!

Nice! I am glad to step you though this process 😄  It is great experience to get (especially so early in your career!). 

(Always feel free to let me know if you want me to take the reigns on the PR – I don't want you to feel like you are committed to doing work if you don't have time. Otherwise I am super happy to continue guiding you step-by-step).

There are two things that we can improve on to make this PR A+++.
 1.  We can "broaden" the type-hint for `file` a bit broader
 2.  MyGrad's codebase abides by certain formatting standards that are enforced using convenient auto-formatters that we can run.
 
 ### Broadening the type-hint for `file`
 You correctly annotated `file: str` in the functions, but users can actually pass some other valid object types for `file`. The `pathlib.Path` type is provided by Python's standard library, and it is a useful object for representing file paths. You can read about them [here](https://www.pythonlikeyoumeanit.com/Module5_OddsAndEnds/WorkingWithFiles.html#pathlib.Path). To make a long story short, users can also pass `pathlib.Path` objects to `file` and they will work great! Thus you can update the annotation for `file` to be:

```python
from pathlib import Path
from typing import Union

def save(file: Union[str, Path], ...)

# (and same for load)
```

if we didn't broaden this type-annotation, then IDEs would incorrectly warn users against passing in a `Path` instance here.

### Auto-formatting code.
MyGrad abides by a code-style standard known as PEP8. You can read about it (and about auto-formatters) [here](https://www.pythonlikeyoumeanit.com/Module5_OddsAndEnds/Writing_Good_Code.html#The-PEP8-Style-Guide-to-Python-Code).

There are two auto-formatters that we use to enforce a uniform code format throughout MyGrad: [isort](https://pycqa.github.io/isort/) and [black](https://black.readthedocs.io/en/stable/).
  - isort manages the unruly stack of import statements that quickly accrue at the top of `.py` files. It will sort import statements alphabetically, and will group them in accordance with PEP8.
  - black is an uncompromising code formatter. You need not spend time formatting code correctly if you run black - it will format your code the same way every time. While this will take care of managing things like line-breaks, indentation, spacing, and brackets for you, it is still your responsibility to write code that is Pythonic and to use good naming conventions.

To install these auto-formatters, run

```shell
pip install isort black
```

And to run them, navigate to the top directory of MyGrad and first run `isort` on all of MyGrad's source code:

```shell
isort src/
```

then run `black`

```shell
black src/
```
These will modify any `.py` files that did not adhere to the standardized styles. You then need to stage/commit any of the modified files and then push the changes. Then, voilà!, all of MyGrad will once again adhere to the same code style. 

Auto-formatters are *super* useful! I write messily-formatted code all of the time and then just write isort/black. I even have these bound to hotkeys in my IDE so I can run them with the press of a few keys.

kw-0:
I'm enjoying this, so I will continue despite being a bit busy. I think I'll be able to commit to tonight?

kw-0:
I'm enjoying this, so I will continue despite being a bit busy. I think I'll be able to commit to tonight?

kw-0:
On second thought, I don't think I can spend more time on this. Thanks though!

kw-0:
I read up on black and isort and they're both really cool! Would you recommend that I use these on other personal projects too, or should they just be used for things others will interact with?

rsokl:
>Would you recommend that I use these on other personal projects too, or should they just be used for things others will interact with?

@kw-0 I absolutely recommend that you use those auto-formatters for all of your projects! They save a lot of time (you don't have to worry about getting spacing/indentation/other formatting right with manual typing) and your code will be nice and tidy and uniform across projects.

BTW I added a few commits onto your branch, mainly to add tests for your code. Because I can't push to your fork, I have to open a new PR that I can push to. But your commits are still attributed to you!

See https://github.com/rsokl/MyGrad/pull/383 for the continued PR.

PR #386, opened 2021-09-04
remove bad example from docs; improve type annotations

rsokl:


src/mygrad/_utils/graph_tracking.py
```diff
@@ -1,15 +1,16 @@
 """
 Provides user interface for suspending computational graph tracking and back-propagation
 """
-from functools import wraps
-from typing import Callable
+from functools import wraps, partial
```
rsokl, line 4:
```suggestion
from functools import wraps
```

PR #393, opened 2021-12-31
NEP 29: Drop Python 3.7 and NumPy 1.18; Add Python 3.10 to CI

rsokl:


rsokl:
MyGrad will continue to support 3.7 up to EOL or until NumPy drops support - whichever happens first.

PR #400, opened 2022-02-06
Improve nnet.initializers docs

davidmascharka:
Fix some ugly formatting. Here's the current doc page ugliness:
img


rsokl (approved):
Blegh. Thanks for dealing with this gross stuff.

PR #441, opened 2024-09-01
update python deps and ci and support numpy 2.0

rsokl:


rsokl:
Gotta:
- Kill versioneer and move to pyproject.toml
- Run black?

